PR #18, opened 2015-02-24
Remove redundant assert

timovwb:
Omitting both the discard and storage arguments should raise a ValueError as the docs say. The code is in place, but was never reached due to this assert statement. This would also give a meaningless assertion error instead of the nice error message.


nornagon:
Good catch. Thanks!

PR #19, opened 2015-02-26
Support more complex input descriptions

nornagon:
Permits the following in the manifest:

```json
{
  "inputs": {
    "parameter_name": {
      "type": "volume",
      "default": "3:microliter"
    }
  }
}
```

### DIFF
--- a/autoprotocol/harness.py
+++ b/autoprotocol/harness.py
@@ -12,7 +12,13 @@ import argparse
 '''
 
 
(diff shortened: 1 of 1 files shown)


therzka:
@nornagon what about description?

nornagon:
@therzka that's also supported, and I'll be adding a few more (e.g. `range`, `required`, `unit`, etc.)

PR #25, opened 2015-03-05
Fix docs typo: methdods -> methods

kyleabeauchamp:
Note that I didn't update the files in `docs/_build/`, as AFAIK the online docs get rebuilt from the source.


therzka:
thanks :+1:

PR #28, opened 2015-03-10
python3 support

naderm:
* Removed use of `.iteritems()` and `.itervalues()`, iterate over dictionary's keys and manually grab the values
* Fall back on `str` type from `basestring`
* Define `xrange = range` when running python3
* Define `__eq__` and `__lt__` for python3 comparisons

The unit tests pass when running Python 3.4.2 as well as Python 2.7.5. I found a few places in the code where python2 conventions were still used, but the tests still passing. I fixed them, but there may yet be other functions where using python3 would throw an error.

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -330,7 +330,10 @@ class Protocol(object):
 
         """
         return {
-            "refs": dict(map(lambda (k, v): (k, v.opts), self.refs.items())),
+            "refs": dict(
+                (k, v.opts)
+                for key, value in self.refs.items()
+            ),
             "instructions": map(lambda x: self._refify(x.data),
                                 self.instructions)
         }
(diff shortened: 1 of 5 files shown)


naderm:
Also, in a most of these places, the code could be somewhat simplified by using the python [six](http://pythonhosted.org//six/) library. It includes functions for iteritems (`six.iteritems(dict)`) / itervalues (`six.itervalues(dict)`), xrange (`six.moves.range()`), and checking string types (`isinstance(obj, six.string_types)`)

I can go back and re-write the changes using that instead, but I wasn't sure if the autoprotocol maintainers would rather avoid the extra dependency.

autoprotocol/protocol.py
```diff
@@ -330,7 +330,10 @@ def as_dict(self):
 
         """
         return {
-            "refs": dict(map(lambda (k, v): (k, v.opts), self.refs.items())),
+            "refs": dict(
+                (k, v.opts)
```
therzka: should be `(key, value.opts)`

naderm: Woops, corrected that.

therzka:
thanks for the python 3 compatible changes, I think we'll stick with these changes and not use six for now but the suggestion is much appreciated :)

PR #41, opened 2015-04-02
added flow_analyze method to Protocol, FlowAnalyze class to Instruction

therzka:


cojofra:
- you changed the order of args in the example to required, optional. but for people reading the example it might be easier to understand how this instruction works if the other is the same as in the asc doc.
- maybe add assert that all `color` in `"minimize_bleed": [{"from": color,"to": [color]}, ...]} ` match a `name` in channel[color]
- maybe in the future we can add helper methods to generate things like area,width,height? like signal(true,true,false) or voltage(356,490). but I think that's probably the extend of making this more elegant/easy to use for now

PR #42, opened 2015-04-09
Unit refactor

wolfram74:
cleaned up unit classes methods, numerous instances where already existing _check_type could have been used but wasn't.


wolfram74:
Forgot to mention this, but available testing suite identified no new errors or failures in with refactors.

therzka:
thanks!

PR #53, opened 2015-05-04
Added tests for procotol.py, util.py, and pipette_tools.py

naderm:
Of note, I made the following changes to the autoprotocol code:

* added `__eq__` operators to Well, WellGroup, and Container.
* Fixed a few small python3 bugs
* Fixed a bug in `Protocol.container_type` when the argument was a ContainerType object
* Added a missing import for `sys` within `pipette_tools.py`
* Made a few whitespace changes for clarity / pep8 compliance


autoprotocol/pipette_tools.py
```diff
@@ -40,7 +45,7 @@ def dispense_target(depth=None, dispense_speed=None, cal_volume=None):
         .. code-block:: json
 
           {
-            "start": "50:microliver/second",
+            "start": "50:microliter/second",
```
therzka: lol good catch, thank you

therzka:
I had to revert this, since it broke some things... if you want you can make another pull request with just the additional tests for now

naderm:
Do you know what were the things that broke? The only changes that I can think of that altered existing behavior were the additions of `__eq__` operators to Container, Well, and WellGroup

PR #55, opened 2015-05-05
fix list of list issue in wells

cojofra:
@therzka allowing lists of wells to be submitted according to spec (or just change description)


therzka:
awesome, thanks @cojofra :+1:

PR #56, opened 2015-05-06
add ability to transfer from mutiple sources to one destination

cojofra:
adjust transfer to take multiple src but only one dst. this is not consolidate since it uses a different tip for every transfer/source well
also: some multiple of 4 indentation clean-up


cojofra:
need to revisit this. breaks some existing stuff. sorry.
will update when fixed

PR #60, opened 2015-05-30
Autopick needs a dataref

ianohara:
The exact form of the data produced by `autopick` is not yet decided, but it will produce data (eg: how many colonies were actually picked?)


therzka:
@ianohara closing this because it was done in a recent commit :)

PR #61, opened 2015-06-03
Sabir sublime text snippets

sabirsaluja:


autoprotocol-python SublimeText Snippets/fill_wells.sublime-snippet
```diff
@@ -0,0 +1,6 @@
+<snippet>
+    <content><![CDATA[
+fill_wells(${1:dst_group}, ${2:src_group}, ${3:volume}, ${4:distribute_target})
```
therzka: this is a private method that doesn't really need a snippet

PR #63, opened 2015-06-12
Make instruction instantiation more concise

peteryin21:
Small change to simplify logic in Protocol constructor.


therzka:
thanks :)

PR #70, opened 2015-07-18
harness: add custom temperature for internal use

ianohara:
Adds the `warm_30` temperature.  This is an oddball temperature that should only be used by those who know they can use it.


ianohara:
Note: explicitly not adding `warm_30` to `instructions.py` and `protocol.py` docs because you probably shouldn't be using `warm_30`.

Issue #72, opened 2015-08-04
Math on Unit quirk


bmiles:
Just making a note of some behaviour with ```Unit``` that caught me out.

```python
In [24]: ben = Unit(5, "microliter")

In [25]: ben * 10
Out[25]: Unit(50.0, microliter)

In [26]: 10 * ben
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-26-91865e2cdf8c> in <module>()
----> 1 10 * ben

TypeError: unsupported operand type(s) for *: 'int' and 'Unit'
```

PR #73, opened 2015-08-04
Support int * Unit

bmiles:
This PR adds support to perform ```int * Unit``` previously only ```Unit * int``` was supported. In reference to #72


therzka:
thanks, Ben... you beat me to it :)

PR #86, opened 2015-11-14
Add acoustic_transfer Instruction class and Protocol method

therzka:


autoprotocol/instruction.py
```diff
@@ -92,6 +92,38 @@ def __init__(self, ref, reagent, columns, speed):
...
+    groups : list(dict)
+        List of `transfer` groups in the form of:
+
+        .. code-block:: json
+
+        {
+            "transfer": [
+                "to": well,
+                "from": well,
+                "volume": volume
```
nornagon: nit: needs `{}`

autoprotocol/protocol.py
```diff
@@ -1069,6 +1069,203 @@ def consolidate(self, sources, dest, volumes, allow_carryover=False,
         else:
             self._pipette([cons])
 
+    def acoustic_transfer(self, source, dest, volume, one_source=False,
+                          droplet_size="0.025:microliter"):
```
nornagon: why not specify in nanoliter?

PR #87, opened 2016-01-27
Bring container.py test coverage to 100%

tdsmith:
Also fixes a couple of minor test and documentation bugs and improves type consistency from some functions (Container.quadrant, Container.inner_wells).

This is my first PR against this project so please let me know if I've missed any style conventions, etc!

### DIFF
--- a/test/container_test.py
+++ b/test/container_test.py
@@ -165,7 +165,7 @@ class ContainerWellGroupConstructionTestCase(unittest.TestCase):
         self.assertEqual(30, len(ws_bigger))
         ws_plus_well = ws + another_container.well(0)
         self.assertEqual(16, len(ws))
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(TypeError):
             ws = ws + "not a well"
 
 class WellVolumeTestCase(unittest.TestCase):
--- a/test/container_test.py
+++ b/test/container_test.py
@@ -111,7 +111,7 @@ class WellVolumeTestCase(unittest.TestCase):
         c.well(1).set_volume(".1:milliliter")
         self.assertTrue(c.well(1).volume == Unit(100, "microliter"))
         with self.assertRaises(ValueError):
-            c.well(2).set_volume("1:liter")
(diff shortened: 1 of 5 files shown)


test/container_test.py
```diff
@@ -111,7 +217,7 @@ def test_set_volume_unit_conv(self):
         c.well(1).set_volume(".1:milliliter")
         self.assertTrue(c.well(1).volume == Unit(100, "microliter"))
         with self.assertRaises(ValueError):
-            c.well(2).set_volume("1:liter")
+            c.well(2).set_volume("1:milliliter")
```
therzka: The old test was correct, please correct this to `self.assertRaises(ValueError, c.well(2).set_volume, "1:liter")`

tdsmith: I think the test wasn't quite right; the volume check in `set_volume` was appearing as skipped in coverage despite this test because it was never being exercised. set_volume _was_ failing with a ValueError, but the ValueError was being emitted from `convert_to_ul` because it doesn't know what a liter is (it understands [nl, µl, and ml](https://github.com/autoprotocol/autoprotocol-python/blob/master/autoprotocol/util.py#L16)).  If you move the "liter" test outside of the `assertRaises` block you'll see that the traceback isn't from the check in `set_volume`.

tdsmith: I wasn't sure whether to prefer the context-manager or method invocation of `assertRaises`; both were already present and I ended up using either arbitrarily. Do you have a preference?

therzka: nope, either way is fine

I see what you mean, we should probably add tests for utils.py

therzka: could you just remove that assertion altogether?  Then I will merge this

tdsmith: Will do. I think it's useful to test this, though; the ml test correctly tests the maximum-volume check in set_volume, since the dummy_type well capacity is 200 µl, so you'll lose coverage if we remove this test.

therzka: Ah I see... nevermind then!

tdsmith:
Cool, thanks!

PR #91, opened 2016-03-10
Add mag transfer

transcripticpll:


autoprotocol/instruction.py
```diff
@@ -70,6 +70,52 @@ def __init__(self, groups):
         })
 
 
+class Magnetic_Transfer(Instruction):
```
therzka: CamelCase

transcripticpll: thanks and fixed

autoprotocol/util.py
```diff
@@ -226,6 +227,36 @@ def check_stamp_append(current_xfer, prev_xfer_list, maxTransfers=3,
     return True
 
 
+def check_valid_mag(container, head):
+    if head == "96-deep":
+        if container.container_type.shortname not in ["96-v-kf", "96-deep-kf", "96-deep", "96-flat", "96-flat-uv"]:
```
therzka: could probably either make these container type lists constants or do the check based on well_count && well_volume_ul

PR #109, opened 2016-06-06
Add time constraints

transcripticpll:
Summary:
Add time constraints top-level field in accordance with the time constraints ASC
https://docs.google.com/document/d/1pbI6OgWu8TV7P6jr9klrR-WwN7qowgWNBSq9KQ7jheo/edit

Fixes T6227

Test Plan: passes tox

Reviewers: yangchoo, Vanessa

Maniphest Tasks: T6227

Differential Revision: https://work.r23s.net/D4951

Improvements include:

 - Add mirroring option

 - Add checks for impossible constraints

 - Add checks for <0 index instructions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/)
 ## Unreleased
 ---
 Added
+- Time constraints
 
 Changed
 - illuminaseq allows cycle specification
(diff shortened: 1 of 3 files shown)


transcripticpll:
This represents version 1 of time constraints which allows for usability. For future improvements, instruction methods on the protocol object could return instruction objects which could also get their position in the protocol.instructions list. These could be used to better mark time constraint start/stops.

PR #136, opened 2016-07-25
Distribute Instruction Fix

polarpine:
Allow `distribute_speed` and properly assign `distribute_target` in the Distribute Instruction.


transcripticpll:
looks good. Thanks!

PR #147, opened 2016-08-21
Update valid incubate_before params

polarpine:
Additional validation on optional `incubate_before` parameters in Absorbance, Luminescence and Fluorescence Instructions.


transcripticpll:
lgtm. thanks!

PR #154, opened 2016-10-24
Dye Test

drjimmypants:
Add ability to use `--dye_test` flag with `transcriptic preview` to convert a run into a water/dye test.


yangchoo:
lgtm

PR #160, opened 2018-06-19
Change default linting engine to pylint

rhysormond:
Pylint allows for much more powerful and configurable linting compared to flake8. In particular this will be really helpful in standardizing docstrings.

### DIFF
--- a/.arclint
+++ b/.arclint
@@ -1,12 +1,10 @@
 {
   "linters": {
-    "flake8-warning": {
-      "type": "flake8",
+    "pylint": {
+      "type": "pylint",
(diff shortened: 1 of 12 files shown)


yangchoo:
Awesome, thanks!

Could you also modify the `.travis.yml` file so that it runs pylint from tox?

Since we're using tox-travis, we'll probably have to do something along the lines of `https://tox-travis.readthedocs.io/en/stable/envlist.html`

rhysormond:
Currently failing because docs don't build properly. Deciding whether we want to move forward with broken docs or fix them as part of this diff.

PR #164, opened 2018-07-24
Update docs/requirements for RTD

yangchoo:
We need to update the docs/requirements.txt in order for ReadTheDocs to build properly

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-Sphinx==1.3
-sphinxcontrib-napoleon
+Sphinx>=1.7,<2
+sphinx-rtd-theme
 releases
 -e git://github.com/autoprotocol/autoprotocol-python.git@master#egg=autoprotocol
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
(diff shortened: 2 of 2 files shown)


docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
-Sphinx==1.3
-sphinxcontrib-napoleon
+Sphinx>=1.7,<2
```
rhysormond: in the future it would be great if we could find some way of versioning this w/ the requirements in setup.py

outside of the scope of this diff though

docs/requirements.txt
```diff
@@ -1,4 +1,4 @@
...
+sphinx-rtd-theme
 releases
--e git://github.com/autoprotocol/autoprotocol-python.git@master#egg=autoprotocol
\ No newline at end of file
+-e git://github.com/autoprotocol/autoprotocol-python.git@master#egg=autoprotocol
```
rhysormond: as far as I can tell these lines are the same. is this just git being weird?

yangchoo: yea.. as far as I can tell, git is weird. I do not intend to change this line and manually examining the diff, it looks . the same

docs/requirements.txt
yangchoo: yup agreed

PR #165, opened 2018-07-24
Fix instruction serialization

rhysormond:
The `op` field has been removed from `Instruction.data`. The correct way to get an `Instruction`'s full specification is now to reference its `__dict__`.

### DIFF
--- a/autoprotocol/instruction.py
+++ b/autoprotocol/instruction.py
@@ -8,7 +8,6 @@ Contains all the Autoprotocol Instruction objects
 """
 
 # pragma pylint: disable=too-few-public-methods, redefined-builtin
-import json
 from functools import reduce
 from .builders import *  # pylint: disable=unused-wildcard-import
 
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`165` update isntruction serialization to use __dict__ as op is no longer included in Instruction data
```
yangchoo: nit: typo `isntruction` -> `instruction`

autoprotocol/instruction.py
```diff
@@ -24,15 +23,18 @@ def __init__(self, op, data):
...
         Returns
         -------
-        str
-            Instruction object encoded as json string
+        dict
+            A JSON serializable representation of the Instruction
+
         """
-        return json.dumps(dict(op=self.op, **self.data), indent=2)
+        return dict(op=self.op, **self.data)
```
yangchoo: hmm... could we still leave in the json method for backwards compatibility? We could wrap it around the `_as_AST` method if we want

rhysormond: going to clean up docs, but as noted in conversation the `Instruction.json` method doesn't actually work when there are non JSON-serializeable objects referenced in its data. the most prevalent case of this is when an Instruction references a Container or other Autoprotocol Python object.

in general it would be nice to move towards more robust (de)serialization in the future.

PR #171, opened 2018-08-18
Add FlowCytometry instruction

dbdalton2000:
Summary: Adds a basic flow cytometry instruction and protocol method.

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -3769,7 +3769,7 @@ class Protocol(object):
 
     def flow_cytometry(self, dataref, samples, lasers,
                        collection_conditions, width_threshold=None,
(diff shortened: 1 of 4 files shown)


autoprotocol/protocol.py
```diff
@@ -3767,6 +3767,154 @@ def uncover(self, ref, store_lid=None):
...
+                        "longpass": "465:nanometer"
+                      },
+                      "detector_gain": "10:millivolt"
+                    }
+                  ]
+                }
+              ],
+              "collection_conditions": {
+                "acquisition_volume": "5:microliter",
+                "flow_rate": "12.5:microliter/minute",
```
rhysormond: I think that for internal consistency we're going to make this `flowrate` even though the accepted spelling seems to be `flow rate`.

PR #177, opened 2018-09-05
update Incubate instruction/protocol co2 parameter docstrings

dbdalton2000:
This PR updates the docstrings for the Incubate instruction and related protocol method

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -2448,7 +2448,7 @@ class Protocol(object):
         shaking : bool, optional
             Specify whether or not to shake container if available at the
             specified temperature
-        co2 : int, optional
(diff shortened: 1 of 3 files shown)


autoprotocol/protocol.py
```diff
@@ -2448,7 +2448,7 @@ def incubate(self, ref, where, duration, shaking=False, co2=0,
         shaking : bool, optional
             Specify whether or not to shake container if available at the
             specified temperature
-        co2 : int, optional
+        co2 : Number, optional
```
yangchoo: thanks for fixing this! This matches up with the original ASC.

To help clarify things for future developers, could you add a type-check below for Number? (so that its clearer that this is in reference to numbers.Number)

PR #188, opened 2019-02-21
Add flag to propagate aliquot props with liha

rhysormond:
Add a flag to `Protocol` to propagate aliquot properties when liquid handling.
Add new methods to `Container` and `ContainerType` for fetching WellGroups.

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -7,9 +7,12 @@ Container, Well, WellGroup objects and associated functions
 
 """
 
-from __future__ import print_function
-from .unit import Unit
+from __future__ import print_function, division
+import warnings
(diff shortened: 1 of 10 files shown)


autoprotocol/container.py
```diff
@@ -54,10 +57,15 @@ def validate_properties(properties):
                 raise TypeError(
                     "Aliquot property {} : {} has a key of type {}, "
                     "it should be a 'str'.".format(key, value, type(key)))
-            if not isinstance(value, basestring):
+            try:
```
yangchoo: Thinking about it... I think this is something we should highlight out in the changelog as a separate line-item as well

PR #193, opened 2019-04-19
Fix for wells_with to add support for non-str values

yangchoo:
We added support for non-str values in properties as part of PR190.  We did not update `wells_with` and this adds in the missing support.


joshhacksthings:
I also want to add a comment here that TX's internal infrastructure (WEB) prohibits the use of anything except for a `String` type as a value for a property...

PR #194, opened 2019-05-30
add seal capability to 384-flat

danielwenwen:


danielwenwen:
@VarunKanwar could you review this and approve this merge? thanks!

rhysormond:
It would be great if you could set the terribly named `peioritize_seal_or_cover` to `cover`for this ContainerType. Historically because this container had no seal capabilities we would always cover it for implicit covers and with this change we would now seal by default.

In the long-term I think we should eliminate these implicit cover/seal steps but for now we should keep this backwards compatible.

See: https://github.com/danielwenwen/autoprotocol-python/blob/53ad9e8de3d27b253325538881e15362f8bfe8eb/autoprotocol/container_type.py#L29

danielwenwen:
@rhysormond made the changes, let me know if there's anything else. I saw that tx also update the changelog, not sure if I should do that though

PR #196, opened 2019-06-05
New harness function for easily getting preview parsed from manifest

joshhacksthings:
Functionality to provide manifest preview in interactive python mode

### DIFF
--- a/autoprotocol/harness.py
+++ b/autoprotocol/harness.py
@@ -26,20 +26,78 @@ _DYE_TEST_RS = {
 }
 
 
-def get_protocol_preview(protocol, name):
-    manifest_json = io.open('manifest.json', encoding='utf-8').read()
+def get_protocol_preview(protocol, name, manifest='manifest.json'):
+    """
+    Parses the 'preview' section of a manifest to use as protocol inputs
+
+        Example Usage:
+
+        .. code-block:: python
+
+          p = Protocol()
+          preview = get_protocol_preview(p, name="qPCR")
+
+        Output:
+
+        .. code-block:: python
+        {
+            "qPCR_input1": "value",
+            "qPCR_input2": "value2",
+            "qPCR_group": {
+                "group_entry1": "value",
+                "group_entry2": "value"
+            }
+        }
+
+
+    Parameters
+    ----------
+    protocol : Protocol
+        Protocol object being parsed.
+    name : str
+        Name of protocol in manifest to get preview for
+    manifest : str, optional
+        Name of manifest file
+
+    Returns
+    -------
+    dict
+        Dictionary of parameters used as a Protocol input
+
+    Raises
+    ------
+    RuntimeError
+        No manifest.json file present in directory
+    RuntimeError
+        Protocol not found in manifest
+    RuntimeError
+        More than one protocol found in manifest
+
+    """
+    try:
+        manifest_json = io.open(manifest, encoding='utf-8').read()
+    except IOError:
+        raise RuntimeError(
+            "'%s' file not found in directory."
+            % manifest
+        )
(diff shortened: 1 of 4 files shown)


joshhacksthings:
hey Strateos team - @VarunKanwar @rhysormond @yangchoo 
just checking if anyone has the bandwidth to review this as this is blocking some of the tooling I'm working on.

autoprotocol/harness.py
```diff
@@ -26,6 +26,23 @@
 }
 
 
+def get_protocol_preview(protocol, name):
```
yangchoo: We haven't historically been good about documentation for things in harness, but would you mind adding some documentation here?

Some blurb about what the function does and an example would be awesome :)

autoprotocol/harness.py
```diff
@@ -26,6 +26,23 @@
...
+    manifest_json = io.open('manifest.json', encoding='utf-8').read()
```
yangchoo: should probably wrap this in a try/catch in case the file doesn't exist

autoprotocol/harness.py
```diff
@@ -26,6 +26,23 @@
...
+    manifest = Manifest(json.loads(manifest_json))
+
+    try:
+        source = [
+            manifest.protocols[i] for i, m in enumerate(manifest.protocols)
+            if manifest.protocols[i]['name'] == name][0]['preview']
```
yangchoo: nit: the formatting is slightly odd here. there's also an implicit picking of the first match, which would be good to call out/document or add an error case for
```
protocols = [
       manifest.protocols[i] for i, m in enumerate(manifest.protocols)
       if manifest.protocols[i]['name'] == name
]
source = protocols[0]['preview'] # or some other check
```

autoprotocol/harness.py
joshhacksthings: can do!

docs/changelog.rst
```diff
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :release:`5.4.2 <2019-06-20>`
```
yangchoo: nit: We usually release as a separate commit. There's no need to include this for now (also, technically this is a not a bugfix, so it should be a minor bump)

test/manifest_test.py
```diff
@@ -605,3 +605,7 @@ def test_cover_type_on_store(self):
         self.protocol.uncover(test2)
         seal_on_store(self.protocol)
         assert (self.protocol.instructions[-1].lid == "low_evaporation")
+
+    def test_get_protocol_preview(self):
+        preview = get_protocol_preview(self.protocol, "TestMethod", manifest="test/manifest_test.json")
+        assert isinstance(preview, dict)
```
yangchoo: Could you match it against the known output instead of just a type check? Thanks!

test/manifest_test.py
```diff
@@ -608,4 +608,8 @@ def test_cover_type_on_store(self):
 
     def test_get_protocol_preview(self):
         preview = get_protocol_preview(self.protocol, "TestMethod", manifest="test/manifest_test.json")
-        assert isinstance(preview, dict)
+        assert "my_string" in preview
```
yangchoo: very minor nit, but could you accumulate these keys into a list and iterate through that instead?

joshhacksthings: Definitely!
So something like:
```
manifest_keys = ["my_string", "my_container", "my_volume", "my_length", "my_bool"]
for key in manifest_keys:
    assert (key in preview)
```

yangchoo: yup, something like that would work

joshhacksthings: great! thanks!

PR #198, opened 2019-07-17
Prep v542

joshhacksthings:
Preparing for release of 5.4.2 which includes my previously added `get_protocol_preview` function in harness.


joshhacksthings:
Great catch Yang! Thanks for the feedback, I've incorporated changes

PR #206, opened 2019-08-20
Upgrade ApPy to Python 3

yangchoo:
With Python 2.7 EOL coming at 1 Jan, 2020, we need to upgrade to Python 3.

As part of this breaking diff, we've also decided to only support Python 3.5 upwards, change the underlying test infrastructure, update and pin some dependencies as well as keeping the documentation up-to-date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 sudo: false
 language: python
 python:
-    - "2.7"
     - "3.5"
+    - "3.6"
+    - "3.7"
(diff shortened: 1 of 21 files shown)


yangchoo:
still to resolve pylint failures

yangchoo:
Fixed a bunch of lint warnings and doc building issues from existing commits that weren't caught by the CI, would be awesome if @joshhacksthings and @polarpine can verify that those are correct

PR #216, opened 2019-10-10
pin pytest-cov package version so docs build

joshhacksthings:


setup.py
```diff
@@ -29,7 +29,7 @@ def run_tests(self):
     'coverage>=4.5, <5',
     'pylint>=1.9, <2',
     'pytest>=4, <5',
-    'pytest-cov>=2, <3',
+    'pytest-cov==2.7.1',
```
rhysormond: is there a reason this needs to be pinned at 2.7.1 instead of, e.g. between 2 and 2.7.1?

joshhacksthings:
I tested versions `2.8.0`, `2.8.1`, found that only `2.8.1` breaks it

PR #221, opened 2019-11-12
Add warm35

joshhacksthings:
Adds `warm_35` as an incubation location


docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`-` Add `warm_37` incubation location
```
polarpine: Should be `warm_35`

PR #223, opened 2020-01-24
Fix well_from_shape bug

yaoyuyang:
well_from_shape was not returning correct wells for index beyond one.
For example, for a 384 well with origin well of A2 and shape of 96 well
to select, it should return the 2nd quadrant, however, the current code
returns the 1st quadrant wells. We added failing tests and fixed this bug
by using a new approach for calculating well indices and checking boundaries.

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -1049,6 +1049,8 @@ class Container(object):
         # coordinates of the tail (bottom right well) should not exceed bounds
         tail_row = well_rows[-1]
         tail_col = well_cols[-1]
+        # tail_row and tail_col are 0-indexed based
+        # container_rows and container_cols are 1-indexed based
         if tail_row + 1 > container_rows or tail_col + 1 > container_cols:
             raise ValueError(
                 "origin: {} with shape: {} exceeds the bounds of "
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -1032,32 +1032,28 @@ class Container(object):
         # getting the row and column values for the origin
         origin_row, origin_col = self.decompose(origin)
 
-        # origin coordinates transformed to SBS format space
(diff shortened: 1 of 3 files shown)


yaoyuyang:
cc @gasnew

test/container_test.py
```diff
@@ -372,6 +372,18 @@ def test_handles_sbs384_containers(self, dummy_384):
             dummy_384.wells_from_shape(0, SHAPE(rows=8, columns=12))
             == dummy_384.quadrant(0)
         )
+        assert(
```
yaoyuyang: These new added tests were failing with the code on the current master. The fix made these tests passing.

autoprotocol/container.py
```diff
@@ -1032,32 +1032,28 @@ def wells_from_shape(self, origin, shape):
...
+            well_row = int(origin_row + idx * row_scaling)
+            well_rows.append(well_row)
+        for idx in range(shape["columns"]):
+            well_col = int(origin_col + idx * col_scaling)
+            well_cols.append(well_col)
+
+        # coordinates of the tail (bottom right well) should not exceed bounds
+        tail_row = well_rows[-1]
+        tail_col = well_cols[-1]
+        if tail_row + 1 > container_rows or tail_col + 1 > container_cols:
```
rhysormond: Ah! this is because tail_row is 0-indexed and container_rows is 1-indexed. Is there any chance you could leave a comment clarifying that?

I was about to suggest `>=` instead of `+ 1 >` but I think your implementation is actually makes the intent clearer.

yaoyuyang: Yes. I can certainly add a comment here!

yaoyuyang: I just added comments above this line. Let me know how it looks to you!

PR #226, opened 2020-02-18
Enable extra detection arguments in position_z dict

hlgirard:
This PR aims at enabling the passing of extra arguments in the `detection` dictionary within the `position_z` dict of the `LiquidHandleMethod` class.

Such extra arguments can be used to handle capacitive / pressure liquid level detection with liquid handlers that do not support the direct passing of thresholds in pascal or farads such as Hamilton STAR class liquid handlers.

The `LiquidHandle.builders.position_z` input checking method was rigidly sanitizing inputs to the expected arguments and stripping any extra arguments passed in the `detection` dict.

### DIFF
--- a/test/liquid_handle_builder_test.py
+++ b/test/liquid_handle_builder_test.py
@@ -122,6 +122,25 @@ class TestLiquidHandleBuilder(object):
             structured_position_z
         )
 
+        extra_detection_arguments_positions_z = {
+            "reference": flat_position_z["reference"],
+            "offset": flat_position_z["offset"],
+            "move_rate": flat_position_z["move_rate"],
+            "detection": {
+                "method": flat_position_z["detection_method"],
+                "threshold": flat_position_z["detection_threshold"],
+                "duration": flat_position_z["detection_duration"],
+                "fallback": flat_position_z["detection_fallback"],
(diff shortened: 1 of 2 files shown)


test/liquid_handle_builder_test.py
```diff
@@ -122,6 +122,25 @@ def test_position_z_builder(self):
...
+        extra_detection_arguments_positions_z = {
+            "reference": flat_position_z["reference"],
+            "offset": flat_position_z["offset"],
+            "move_rate": flat_position_z["move_rate"],
+            "detection": {
+                "method": flat_position_z["detection_method"],
+                "threshold": flat_position_z["detection_threshold"],
+                "duration": flat_position_z["detection_duration"],
+                "fallback": flat_position_z["detection_fallback"],
+                "x_edge_steepness": 20,
```
rhysormond: nit: I'd prefer avoiding referencing specific `x_`params here and instead throwing some other field here that's obviously out of the ordinary (e.g. `foo`, `test` or similar).

This is partly because we don't want those specific fields names to seem important to this test and partly because we'd like to keep the notion of `x_`params out of autoprotocol python as much as possible.

PR #229, opened 2020-04-01
Updating LiquidHandle instruction with 'density' parameter and 'positive_displacement' mode

aota001:
The viscosity of liquid samples used in a laboratory ranges from very fluid (ie. acetone) to very thick (ie. glycerol). A limitation to the current `liquid_handle` instruction is that there is only an `air-displacement` mode to handle water-like liquid. While this is sufficient for liquids with relatively low viscosity, they should not be used for viscous samples due to the inability for air pressure to overcome the surface tension. 

To provide a more accurate way to handle these samples, the `positive-displacement` mode should be used for 'viscous' liquid class. Furthermore, there are devices that are capable of requesting liquid transfer amount in mass instead of volume. To allow such feature, optional density parameter is added to be able to convert mass to volume (and vice versa) as needed.

Finally, these changes only add optional parameter, and does not affect the current default behavior of the LiuqidHandle instruction. 'air_displacement' mode is used for all liquids that are liquid_class not indicated (None), 'default' or 'air' as previously done.

### DIFF
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1305,7 +1305,7 @@ class LiquidHandleBuilders(InstructionBuilders):
             "transports": transports
         }
 
-    def transport(self, volume=None, pump_override_volume=None,
+    def transport(self, volume=None, density=None, pump_override_volume=None,
                   flowrate=None, delay_time=None, mode_params=None):
         """Helper for building transports
 
@@ -1314,6 +1314,8 @@ class LiquidHandleBuilders(InstructionBuilders):
         volume : Unit or str, optional
             Volume to be aspirated/dispensed. Positive volume -> Dispense.
             Negative -> Aspirate
+        density : Unit or str, optional
+            Density of the liquid to be aspirated/dispensed.
         pump_override_volume : Unit or str, optional
             Calibrated volume, volume which the pump will move
         flowrate : dict, optional
@@ -1333,6 +1335,8 @@ class LiquidHandleBuilders(InstructionBuilders):
         """
         if volume is not None:
             volume = parse_unit(volume, "ul")
+        if density is not None:
+            density = parse_unit(density, "mg/ml")
         if pump_override_volume is not None:
             pump_override_volume = parse_unit(pump_override_volume, "ul")
         if flowrate is not None:
@@ -1344,6 +1348,7 @@ class LiquidHandleBuilders(InstructionBuilders):
 
         return {
             "volume": volume,
+            "density": density,
             "pump_override_volume": pump_override_volume,
             "flowrate": flowrate,
             "delay_time": delay_time,
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1251,12 +1251,13 @@ class LiquidHandleBuilders(InstructionBuilders):
     """
     def __init__(self):
         super(LiquidHandleBuilders, self).__init__()
-        self.liquid_classes = ["air", "default"]
+        self.liquid_classes = ["air", "default", "viscous"]
         self.xy_max = 1
         self.z_references = [
             "well_top", "well_bottom", "liquid_surface", "preceding_position"
         ]
         self.z_detection_methods = ["capacitance", "pressure", "tracked"]
+        self.dispense_modes = ["air_displacement", "positive_displacement"]
 
     def location(self, location=None, transports=None):
         """Helper for building locations
@@ -1405,7 +1406,7 @@ class LiquidHandleBuilders(InstructionBuilders):
 
         Parameters
         ----------
-        liquid_class : Enum({"default", "air"}), optional
+        liquid_class : Enum({"default", "air", "viscous"}), optional
             The name of the liquid class to be handled. This affects how
             vendors handle populating liquid handling defaults.
         position_x : dict, optional
@@ -1742,6 +1743,72 @@ class LiquidHandleBuilders(InstructionBuilders):
             "initial_z": initial_z,
             "flowrate": flowrate
         }
+
+    def mode(self, transports=None, mode=None):
+        """Helper for selecting dispense mode based on liquid_class name
+
+        Parameters
+        ----------
+        trasnports : dict, optional
+            Dictionary of the transport parameters
+        mode : str, optional
+            Mode of dispense type
+
+        Returns
+        -------
+        str
+            Mode of dispense type
+
+        Raises
+        ------
+        TypeError
+        ValueError
+
+        """
+        liquid_classes = set([transport["mode_params"]["liquid_class"] for transport in transports])
+        # remove automatically added 'air' class from blowout, etc.
+        non_air_classes = [liq for liq in liquid_classes if liq != "air"]
+        if mode:
+            if mode not in self.dispense_modes:
+                raise ValueError(
+                    "mode: {} must be one of the valid modes: {}"
+                    "".format(mode, self.dispense_modes)
+                )
+        modes = []
+        # get mode for each liquid_class if not specified already from the user input.
+        for liquid in non_air_classes:
+            if liquid is not None:
+                if not isinstance(liquid, str):
+                    raise TypeError(
+                        "liquid: {} is not of type str or None".format(liquid)
+                    )
+                if liquid not in self.liquid_classes:
+                    raise ValueError(
+                        "liquid_class: {} must be one of the valid classes: {} "
+                        "".format(liquid, self.liquid_classes)
+                    )
+                if mode is None:
+                    if liquid == "viscous":
+                        modes.append("positive_displacement")
+                    else:
+                        modes.append("air_displacement")
+            else:
+                if mode is None:
+                    modes.append("air_displacement")
+        # return error if there are incompatible liquid_class in one set of transports.
+        if len(set(modes)) > 1:
+            raise ValueError(
+                "There are multiple liquid classes which could potentially have different modes: {}."
+                "Please specify the mode to be used from: {}."
+                "".format(modes, self.dispense_modes)
+            )
+        # if all liquid classes happen to be "air", mode is "air_displacement"
+        elif len(modes) == 0:
+            mode = "air_displacement"
+        else:
+            mode = list(modes)[0]
+
+        return mode
 
 
 class PlateReaderBuilders(InstructionBuilders):
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1846,7 +1846,8 @@ class LiquidHandleBuilders(InstructionBuilders):
             # remove automatically added 'air' class from blowout, etc.
             non_air_classes = [liq for liq in liquid_classes if liq != "air"]
             # if liquid classes only contain "air" type, mode is returned.
-            if not non_air_classes:
+            has_only_air_classes = not non_air_classes
+            if has_only_air_classes:
                 return liquid_class_to_dispense_modes["air"]
 
             modes = []
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1236,7 +1236,7 @@ class SpectrophotometryBuilders(InstructionBuilders):
             return self.position_z_calculated(
                 **position_z["calculated_from_wells"]
             )
-        if "manual" in position_z:
+        elif "manual" in position_z:
             return self.position_z_manual(
                 **position_z["manual"]
             )
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -24,7 +24,6 @@ Instruction
     Instructions corresponding to each of the builders
 """
 
-#pylint: disable=C0412
 from collections import defaultdict
 from numbers import Number
 
@@ -654,7 +653,6 @@ class DispenseBuilders(InstructionBuilders):
 
 
 class SpectrophotometryBuilders(InstructionBuilders):
-    # pylint: disable=C0103
     """
     These builders are meant for helping to construct arguments for the
     `Spectrophotometry` instruction.
@@ -1309,8 +1307,8 @@ class LiquidHandleBuilders(InstructionBuilders):
             "transports": transports
         }
 
-    def transport(self, volume=None, density=None, pump_override_volume=None,
-                  flowrate=None, delay_time=None, mode_params=None):
+    def transport(self, volume=None, pump_override_volume=None, flowrate=None,
+                  delay_time=None, mode_params=None, density=None):
         """Helper for building transports
 
         Parameters
@@ -1318,8 +1316,6 @@ class LiquidHandleBuilders(InstructionBuilders):
         volume : Unit or str, optional
             Volume to be aspirated/dispensed. Positive volume -> Dispense.
             Negative -> Aspirate
-        density : Unit or str, optional
-            Density of the liquid to be aspirated/dispensed.
         pump_override_volume : Unit or str, optional
             Calibrated volume, volume which the pump will move
         flowrate : dict, optional
@@ -1330,6 +1326,8 @@ class LiquidHandleBuilders(InstructionBuilders):
         mode_params : dict, optional
             Mode parameters
             See Also LiquidHandle.builders.mode_params
+        density : Unit or str, optional
+            Density of the liquid to be aspirated/dispensed in mg/ml.
 
 
         Returns
@@ -1337,17 +1335,17 @@ class LiquidHandleBuilders(InstructionBuilders):
         dict
             transport parameters for a LiquidHandle instruction
         """
-        if volume is not None:
+        if volume:
             volume = parse_unit(volume, "ul")
-        if density is not None:
+        if density:
             density = parse_unit(density, "mg/ml")
-        if pump_override_volume is not None:
+        if pump_override_volume:
             pump_override_volume = parse_unit(pump_override_volume, "ul")
-        if flowrate is not None:
+        if flowrate:
             flowrate = self.flowrate(**flowrate)
-        if delay_time is not None:
+        if delay_time:
             delay_time = parse_unit(delay_time, "s")
-        if mode_params is not None:
+        if mode_params:
             mode_params = self.mode_params(**mode_params)
 
         return {
@@ -1409,7 +1407,7 @@ class LiquidHandleBuilders(InstructionBuilders):
 
         Parameters
         ----------
-        liquid_class : Enum({"default", "air", "viscous"}), optional
+        liquid_class : str, optional
             The name of the liquid class to be handled. This affects how
             vendors handle populating liquid handling defaults.
         position_x : dict, optional
@@ -1747,8 +1745,52 @@ class LiquidHandleBuilders(InstructionBuilders):
             "flowrate": flowrate
         }
 
-    def mode(self, transports=None, mode=None):
+    def desired_mode(self, transports=None, mode=None):
         """Helper for selecting dispense mode based on liquid_class name
+        For non-viscous, water-like liquid and air, the method will default
+        to "air_displacement". To allow for more accurate aspirate and
+        dispense volumes, 'viscous' liquid where air pressure alone is
+        sometimes not sufficient to overcome the surface tenstion to pull
+        or push the liquid sufficiently through transfer tip, it will
+        default to "positive_displacement" unless otherwise specified.
+
+        Example Usage:
+        .. code-block:: python
+        example_transports = [
+            LiquidHandle.builders.transport(
+                volume=Unit(1, "uL"),
+                density=None,
+                pump_override_volume=Unit(2, "uL"),
+                flowrate=LiquidHandle.builders.flowrate(
+                    target=Unit(10, "uL/s")),
+                delay_time=Unit(0.5, "s"),
+                mode_params=LiquidHandle.builders.mode_params(
+                    liquid_class="air",
+                    position_z=LiquidHandle.builders.position_z(
+                        reference="preceding_position")
+                )
+            ),
+            LiquidHandle.builders.transport(
+                volume=Unit(1, "uL"),
+                density=None,
+                pump_override_volume=Unit(2, "uL"),
+                flowrate=LiquidHandle.builders.flowrate(
+                    target=Unit(10, "uL/s")),
+                delay_time=Unit(0.5, "s"),
+                mode_params=LiquidHandle.builders.mode_params(
+                    liquid_class="viscous",
+                    position_z=LiquidHandle.builders.position_z(
+                        reference="preceding_position")
+                )
+            )
+        ]
+
+        LiquidHandle.builders.desired_mode(example_transports, None)
+
+        This will return "positive_displacement" based on the "viscous" liquid
+        class. Note that "air" (which defaults to "air_displacement") does not
+        cause a conflict since "air" is often applied in additional transfers
+        of air (such as blowout) for any liquid class, and is disregarded.
 
         Parameters
         ----------
@@ -1773,52 +1815,69 @@ class LiquidHandleBuilders(InstructionBuilders):
         ValueError
             multiple liquid_class exists in one LiquidHandle
         """
-        liquid_classes = set([transport["mode_params"]["liquid_class"]
-                              for transport in transports])
-        # remove automatically added 'air' class from blowout, etc.
-        non_air_classes = [liq for liq in liquid_classes if liq != "air"]
+        # dict for default dispense mode for each liquid_class
+        liquid_class_to_dispense_modes = {
+            "air": "air_displacement",
+            "default": "air_displacement",
+            "viscous": "positive_displacement"
+        }
         if mode:
             if mode not in self.dispense_modes:
                 raise ValueError(
                     "mode: {} must be one of the valid modes: {}"
                     "".format(mode, self.dispense_modes)
                 )
-        modes = []
-        # get mode for each liquid_class if not specified already from the
-        # user input.
-        for liquid in non_air_classes:
-            if liquid is not None:
-                if not isinstance(liquid, str):
-                    raise TypeError(
-                        "liquid: {} is not of type str or None".format(liquid)
-                    )
-                if liquid not in self.liquid_classes:
-                    raise ValueError(
-                        "liquid_class: {} must be one of the valid classes: {} "
-                        "".format(liquid, self.liquid_classes)
-                    )
-                if mode is None:
-                    if liquid == "viscous":
-                        modes.append("positive_displacement")
-                    else:
-                        modes.append("air_displacement")
-            else:
-                if mode is None:
-                    modes.append("air_displacement")
-        # return error if there are incompatible liquid_class in one set of
-        # transports.
-        if len(set(modes)) > 1:
-            raise ValueError(
-                "There are multiple liquid classes which could potentially"
-                "have different modes: {}."
-                "Please specify the mode to be used from: {}."
-                "".format(modes, self.dispense_modes)
-            )
-        # if all liquid classes happen to be "air", mode is "air_displacement"
-        if not modes:
-            mode = "air_displacement"
         else:
-            mode = list(modes)[0]
+            # get liquid_classes from transport input
+            liquid_classes = set([transport["mode_params"]["liquid_class"]
+                                  for transport in transports])
+            # remove automatically added 'air' class from blowout, etc.
+            non_air_classes = [liq for liq in liquid_classes if liq != "air"]
+            # if liquid classes only contain "air" type, mode is returned.
+            if not non_air_classes:
+                return liquid_class_to_dispense_modes["air"]
+
+            modes = []
+            # get mode for each liquid_class if not specified already from the
+            # user input.
+            for liquid in non_air_classes:
+                # resolve mode for different liquid classes (except for None).
+                if liquid:
+                    if not isinstance(liquid, str):
+                        raise TypeError(
+                            "liquid: {} is not of type str".format(liquid)
+                        )
+                    if liquid not in self.liquid_classes:
+                        raise ValueError(
+                            "liquid_class: {} must be one of the valid"
+                            "classes: {} ".format(liquid, self.liquid_classes)
+                        )
+                    if liquid not in liquid_class_to_dispense_modes.keys():
+                        raise ValueError(
+                            "modes: {} did not resolve accordingly. If there "
+                            "is a new liquid_class, make sure dictionary: {}"
+                            "is updated.".format(
+                                set(modes), liquid_class_to_dispense_modes
+                            )
+                        )
+                    modes.append(liquid_class_to_dispense_modes[liquid])
+
+            # if liquid class only contains "air" and/or None, default to
+            # "air_displacement".
+            if not modes:
+                modes.append("air_displacement")
+            # return error if there are incompatible liquid_class a set of
+            # transports.
+            if len(set(modes)) > 1:
+                raise ValueError(
+                    "There are multiple liquid classes which could potentially"
+                    "have different modes: {}."
+                    "Please specify the mode to be used from: {}. Only one mode"
+                    "is allowed per transfer."
+                    "".format(set(modes), self.dispense_modes)
+                )
+            if len(set(modes)) == 1:
+                mode = list(modes)[0]
 
         return mode
 
@@ -1917,7 +1976,6 @@ class EvaporateBuilders(InstructionBuilders):
         self.blowdown_params = ["gas", "blow_rate", "vortex_speed"]
 
     def get_mode_params(self, mode, mode_params):
-        #pylint: disable=R0912
         """
         Checks on the validity of mode and mode_params, and
         creates a dictionary for mode_params
@@ -1977,22 +2035,22 @@ class EvaporateBuilders(InstructionBuilders):
             "spin_acceleration": "g"
         }
         mode_param_output = {}
-        speed_param = [key for key in mode_params.keys()
-                       if key in speed_unit_dict.keys()]
+        speed_param = [(k) for k in mode_params.keys()
+                       if k in speed_unit_dict.keys()]
         if len(speed_param) > 1:
             raise TypeError(
                 "There are multiple speed parameters: {}."
                 "".format(speed_param)
             )
 
-        for speed in speed_param:
-            if Unit(mode_params[speed]).magnitude <= 0:
+        for s in speed_param:
+            if Unit(mode_params[s]).magnitude <= 0:
                 raise ValueError(
-                    "{} is less than or equal to 0.".format(speed)
+                    "{} is less than or equal to 0.".format(s)
                 )
             else:
-                mode_param_output[speed] = parse_unit(
-                    mode_params[speed], speed_unit_dict[speed])
+                mode_param_output[s] = parse_unit(
+                    mode_params[s], speed_unit_dict[s])
 
         if "vacuum_pressure" in mode_params.keys():
             pressure = mode_params["vacuum_pressure"]
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1762,12 +1762,14 @@ class LiquidHandleBuilders(InstructionBuilders):
                 density=None,
                 pump_override_volume=Unit(2, "uL"),
                 flowrate=LiquidHandle.builders.flowrate(
-                    target=Unit(10, "uL/s")),
+                    target=Unit(10, "uL/s")
+                ),
                 delay_time=Unit(0.5, "s"),
                 mode_params=LiquidHandle.builders.mode_params(
                     liquid_class="air",
                     position_z=LiquidHandle.builders.position_z(
-                        reference="preceding_position")
+                        reference="preceding_position"
+                    )
                 )
             ),
             LiquidHandle.builders.transport(
@@ -1775,12 +1777,14 @@ class LiquidHandleBuilders(InstructionBuilders):
                 density=None,
                 pump_override_volume=Unit(2, "uL"),
                 flowrate=LiquidHandle.builders.flowrate(
-                    target=Unit(10, "uL/s")),
+                    target=Unit(10, "uL/s")
+                ),
                 delay_time=Unit(0.5, "s"),
                 mode_params=LiquidHandle.builders.mode_params(
                     liquid_class="viscous",
                     position_z=LiquidHandle.builders.position_z(
-                        reference="preceding_position")
+                        reference="preceding_position"
+                    )
                 )
             )
         ]
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1877,7 +1877,7 @@ class LiquidHandleBuilders(InstructionBuilders):
             # if liquid class only contains "air" and/or None, default to
             # "air_displacement".
             if not modes:
-                modes.append("air_displacement")
+                modes.append(liquid_class_to_dispense_modes["air"])
             # return error if there are incompatible liquid_class a set of
             # transports.
             if len(set(modes)) > 1:
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1336,17 +1336,17 @@ class LiquidHandleBuilders(InstructionBuilders):
         dict
             transport parameters for a LiquidHandle instruction
         """
-        if volume:
+        if volume is not None:
             volume = parse_unit(volume, "ul")
-        if density:
+        if density is not None:
             density = parse_unit(density, "mg/ml")
-        if pump_override_volume:
+        if pump_override_volume is not None:
             pump_override_volume = parse_unit(pump_override_volume, "ul")
-        if flowrate:
+        if flowrate is not None:
             flowrate = self.flowrate(**flowrate)
-        if delay_time:
+        if delay_time is not None:
             delay_time = parse_unit(delay_time, "s")
-        if mode_params:
+        if mode_params is not None:
             mode_params = self.mode_params(**mode_params)
 
         return {
@@ -1831,10 +1831,6 @@ class LiquidHandleBuilders(InstructionBuilders):
             "viscous": "positive_displacement"
         }
         if mode:
-            if not isinstance(mode, str):
-                raise TypeError(
-                    "mode: {} must be in str".format(mode)
-                )
             if mode not in self.dispense_modes:
                 raise ValueError(
                     "mode: {} must be one of the valid modes: {}"
@@ -1872,7 +1868,12 @@ class LiquidHandleBuilders(InstructionBuilders):
                     )
                 )
             modes.add(liquid_class_to_dispense_mode[other_class])
-
+        # return error if modes is empty.
+        if not modes:
+            raise ValueError(
+                "modes: {} resulted in an empty set. Make sure valid mode is "
+                "added for each liquid class."
+            ).format(modes)
         # return error if there are incompatible liquid_class a set of
         # transports.
         if len(set(modes)) > 1:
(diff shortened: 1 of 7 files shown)


autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+        ValueError
+            mode does not contain valid mode value
+        TypeError
+            liquid is not of str
+        ValueError
+            liquid_class is not a valid value
+        ValueError
+            multiple liquid_class exists in one LiquidHandle
+        """
+        liquid_classes = set([transport["mode_params"]["liquid_class"]
```
dzleidig, line 1776: Why not define a map of liquid class to dispense type, e.g.
```
liquid_class_to_dispense_modes = {
		"air" : "air_displacement",
		"default" : "positive_displacement",
		"viscous" : "positive_displacement"
	}

liquid_classes = {"air", "default", "viscous"}

dispense_modes = ["air_displacement", "positive_displacement"]

given_liquid_classes = ["air","default","air","viscous","default","air"]

modes = set(map(lambda given_liquid_class: liquid_class_to_dispense_modes[given_liquid_class], given_liquid_classes))

# modes == {'positive_displacement', 'air_displacement'}
```

autoprotocol/builders.py
```diff
@@ -1251,12 +1254,13 @@ class LiquidHandleBuilders(InstructionBuilders):
     """
     def __init__(self):
         super(LiquidHandleBuilders, self).__init__()
-        self.liquid_classes = ["air", "default"]
+        self.liquid_classes = ["air", "default", "viscous"]
```
dzleidig, line 1257: This can be a set.

autoprotocol/builders.py
```diff
@@ -1251,12 +1254,13 @@ class LiquidHandleBuilders(InstructionBuilders):
...
         self.xy_max = 1
         self.z_references = [
             "well_top", "well_bottom", "liquid_surface", "preceding_position"
         ]
         self.z_detection_methods = ["capacitance", "pressure", "tracked"]
+        self.dispense_modes = ["air_displacement", "positive_displacement"]
```
dzleidig, line 1263: This can be a set.

autoprotocol/builders.py
```diff
@@ -329,7 +330,7 @@ def group(self, steps, cycles=1):
 
         if cycles <= 0:
             raise ValueError("`cycles` {} has to be positive".format(cycles))
-        if len(steps) <= 0:
+        if not steps:
```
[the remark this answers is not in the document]
joshhacksthings: I think this is fine, as she's switching from a numeric comparison of the _length_ of the list to a truthy comparison (based on if there are any items in it).

So the comparison would be:
```
>>> not []
True
>>> []
[]
```

autoprotocol/builders.py
joshhacksthings: To argue in favor of backwards compatibility, this should remain a list.
While I agree a set would be useful here to limit duplicates, I do not think it's a good idea.

autoprotocol/builders.py
```diff
@@ -1333,6 +1339,8 @@ def transport(self, volume=None, pump_override_volume=None,
         """
         if volume is not None:
```
joshhacksthings, line 1340: can we refactor this to `if volume:`

autoprotocol/builders.py
```diff
@@ -1333,6 +1339,8 @@ def transport(self, volume=None, pump_override_volume=None,
...
             volume = parse_unit(volume, "ul")
+        if density is not None:
```
joshhacksthings, line 1342: and this should be `if density`

autoprotocol/builders.py
```diff
@@ -1333,6 +1339,8 @@ def transport(self, volume=None, pump_override_volume=None,
...
+            density = parse_unit(density, "mg/ml")
         if pump_override_volume is not None:
```
joshhacksthings, line 1344: and this should be `if pump_override_volume`

autoprotocol/builders.py
```diff
@@ -1333,6 +1339,8 @@ def transport(self, volume=None, pump_override_volume=None,
...
             pump_override_volume = parse_unit(pump_override_volume, "ul")
         if flowrate is not None:
```
joshhacksthings, line 1346: and `if flowrate:`

autoprotocol/builders.py
```diff
@@ -1400,7 +1409,7 @@ def mode_params(self, liquid_class=None, position_x=None, position_y=None,
 
         Parameters
         ----------
-        liquid_class : Enum({"default", "air"}), optional
+        liquid_class : Enum({"default", "air", "viscous"}), optional
```
joshhacksthings, line 1412: I'd argue calling these an enum, when it is in fact a list, is misleading (enums are a python builtin as of 3.4 iirc). I think we should specify this is a list.

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
             "flowrate": flowrate
         }
 
+    def mode(self, transports=None, mode=None):
```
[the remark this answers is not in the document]
joshhacksthings: I'm also with Dylan here. I had a hard time trying to figure out what `mode` should default to until I realize that (if I understand correctly), this function should return what mode we should be using. As such, it shouldn't include a `mode` argument

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                              for transport in transports])
+        # remove automatically added 'air' class from blowout, etc.
+        non_air_classes = [liq for liq in liquid_classes if liq != "air"]
+        if mode:
+            if mode not in self.dispense_modes:
+                raise ValueError(
+                    "mode: {} must be one of the valid modes: {}"
+                    "".format(mode, self.dispense_modes)
+                )
+        modes = []
```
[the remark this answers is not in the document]
joshhacksthings: ignore this comment. I misread `modes` as `mode`.

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                        "liquid_class: {} must be one of the valid classes: {} "
+                        "".format(liquid, self.liquid_classes)
+                    )
+                if mode is None:
+                    if liquid == "viscous":
+                        modes.append("positive_displacement")
+                    else:
+                        modes.append("air_displacement")
+            else:
+                if mode is None:
```
joshhacksthings, line 1806: this can be rewritten as `if not mode:`

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                    modes.append("air_displacement")
+        # return error if there are incompatible liquid_class in one set of
+        # transports.
+        if len(set(modes)) > 1:
```
joshhacksthings, line 1810: this conditional checks if there is no more than 1 unique mode in `modes`.

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            raise ValueError(
+                "There are multiple liquid classes which could potentially"
+                "have different modes: {}."
+                "Please specify the mode to be used from: {}."
```
joshhacksthings, line 1814: this error message tells the user to specify a mode from a list of modes, but does not say that they should only specify one of these types

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                "".format(modes, self.dispense_modes)
+            )
+        # if all liquid classes happen to be "air", mode is "air_displacement"
+        if not modes:
+            mode = "air_displacement"
+        else:
+            mode = list(modes)[0]
```
joshhacksthings, line 1821: I'd add a few more conditionals to check that the 0th item in this list really is the mode you want. I think including this. Maybe moving this closer to the conditional where you check the length of set of modes is 1 would be a good place for it, then `if len(set(modes)) > 1`, raise that error?

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+        # transports.
+        if len(set(modes)) > 1:
+            raise ValueError(
+                "There are multiple liquid classes which could potentially"
+                "have different modes: {}."
+                "Please specify the mode to be used from: {}."
+                "".format(modes, self.dispense_modes)
+            )
+        # if all liquid classes happen to be "air", mode is "air_displacement"
+        if not modes:
```
joshhacksthings, line 1818: I feel like this could have been determined (and returned) earlier. just a note

autoprotocol/builders.py
```diff
@@ -1833,6 +1917,7 @@ def __init__(self):
         self.blowdown_params = ["gas", "blow_rate", "vortex_speed"]
 
     def get_mode_params(self, mode, mode_params):
+        #pylint: disable=R0912
```
joshhacksthings, line 1920: why are we changing formatting on code not included in this PR? If you want to reformat code not in scope of this PR, please open a new PR

autoprotocol/builders.py
```diff
@@ -1892,22 +1977,22 @@ def get_mode_params(self, mode, mode_params):
             "spin_acceleration": "g"
         }
         mode_param_output = {}
-        speed_param = [(k) for k in mode_params.keys()
-                       if k in speed_unit_dict.keys()]
+        speed_param = [key for key in mode_params.keys()
```
joshhacksthings, line 1980: open a separate PR to commit refactoring

autoprotocol/builders.py
```diff
@@ -1892,22 +1977,22 @@ def get_mode_params(self, mode, mode_params):
...
+                       if key in speed_unit_dict.keys()]
         if len(speed_param) > 1:
             raise TypeError(
                 "There are multiple speed parameters: {}."
                 "".format(speed_param)
             )
 
-        for s in speed_param:
-            if Unit(mode_params[s]).magnitude <= 0:
+        for speed in speed_param:
```
joshhacksthings, line 1988: see above on refactoring

autoprotocol/builders.py
```diff
@@ -1892,22 +1977,22 @@ def get_mode_params(self, mode, mode_params):
...
+            if Unit(mode_params[speed]).magnitude <= 0:
                 raise ValueError(
-                    "{} is less than or equal to 0.".format(s)
+                    "{} is less than or equal to 0.".format(speed)
```
joshhacksthings, line 1991: see above on refactoring

autoprotocol/builders.py
```diff
@@ -1892,22 +1977,22 @@ def get_mode_params(self, mode, mode_params):
...
                 )
             else:
-                mode_param_output[s] = parse_unit(
-                    mode_params[s], speed_unit_dict[s])
+                mode_param_output[speed] = parse_unit(
+                    mode_params[speed], speed_unit_dict[speed])
```
joshhacksthings, line 1995: see above on refactoring

autoprotocol/liquid_handle/liquid_handle_method.py
```diff
@@ -320,7 +320,7 @@ def _is_single_channel(self):
     def _aspirate_simple(self, volume, initial_z,
                          position_x=None, position_y=None,
                          calibrated_vol=None, flowrate=None, delay_time=None,
-                         liquid_class=None):
+                         liquid_class=None, density=None):
```
joshhacksthings, line 323: excellent! glad density is last here. good for backwards compatibility

autoprotocol/protocol.py
```diff
@@ -287,6 +286,7 @@ def ref(self, name, id=None, cont_type=None, storage=None, discard=None,
         )
         self.refs[name] = Ref(name, opts, container)
         return container
+
```
joshhacksthings, line 289: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -490,8 +490,8 @@ def add_time_constraint(self, from_dict, to_dict, less_than=None,
 
         def add_time_constraint_internal(time_const):
             setattr(self, "time_constraints", (
-                getattr(self, "time_constraints", []) + [time_const])
-            )
+                    getattr(self, "time_constraints", []) + [time_const])
```
joshhacksthings, line 493: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -1106,7 +1106,7 @@ def acoustic_transfer(self, source, dest, volume, one_source=False,
             raise RuntimeError(
                 "Transfer volume has to be a multiple of the droplet size ({})."
                 "This is not true for the following volumes: {}"
-                .format(droplet_size, vol_errors)
+                    .format(droplet_size, vol_errors)
```
joshhacksthings, line 1109: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -1363,7 +1363,7 @@ def illuminaseq(self, flowcell, lanes, sequencer, mode, index, library_size,
                     "Illumina sequencing lanes must be a list(dict)"
                 )
             if not all(k in l.keys() for k in [
-                    "object", "library_concentration"]):
+                "object", "library_concentration"]):
```
joshhacksthings, line 1366: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -1385,7 +1385,7 @@ def illuminaseq(self, flowcell, lanes, sequencer, mode, index, library_size,
                              " modes: {}.You specified: {}."
                              "".format(sequencer,
                                        ', '.join(valid_sequencers[
-                                                 sequencer]["modes"]),
+                                                     sequencer]["modes"]),
```
joshhacksthings, line 1388: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -1533,7 +1533,7 @@ def sangerseq(self, cont, wells, dataref, type="standard", primer=None):
 
         if not isinstance(wells, list):
             raise TypeError("Unknown input. SangerSeq wells accepts either a"
-                             "Well, a WellGroup, or a list of well indices")
+                            "Well, a WellGroup, or a list of well indices")
```
joshhacksthings, line 1536: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -1765,7 +1765,7 @@ def dispense(self, ref, reagent, columns, is_resource_id=False,
 
             # Volume accounting
             total_vol_dispensed = (
-                sum([Unit(c["volume"]) for c in columns]) * row_count)
+                    sum([Unit(c["volume"]) for c in columns]) * row_count)
```
joshhacksthings, line 1768: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -4212,7 +4212,7 @@ def flow_analyze(self, dataref, FSC, SSC, neg_controls, samples,
                 raise ValueError("Each sample or control must indicate a "
                                  "volume of type unit. %s" % e)
             if (s.get("captured_events") and not
-                    isinstance(s.get("captured_events"), int)):
+            isinstance(s.get("captured_events"), int)):
```
joshhacksthings, line 4215: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -4221,7 +4221,7 @@ def flow_analyze(self, dataref, FSC, SSC, neg_controls, samples,
                                 "indicating the colors/channels that this"
                                 " control is to be used for.")
             if (c.get("minimize_bleed") and not
-                    isinstance(c.get("minimize_bleed"), list)):
+            isinstance(c.get("minimize_bleed"), list)):
```
joshhacksthings, line 4224: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -4275,7 +4275,7 @@ def flow_analyze(self, dataref, FSC, SSC, neg_controls, samples,
                                         "type dict.")
                     else:
                         if (not c.get("name") or not
-                                isinstance(c.get("name"), str)):
+                        isinstance(c.get("name"), str)):
```
joshhacksthings, line 4278: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -5507,7 +5506,7 @@ def image(self, ref, mode, dataref, num_images=1, backlighting=None,
             raise TypeError("num_images must be a positive integer.")
         if magnification:
             if not isinstance(magnification, (float, int)) or \
-               magnification <= 0:
+                    magnification <= 0:
```
joshhacksthings, line 5509: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -5599,10 +5598,10 @@ def _add_mag(self, sub_op, head, new_tip, new_instruction, sub_op_name):
...
         last_instruction = self.instructions[-1] if self.instructions else None
         maybe_same_instruction = (
-            new_instruction is False and
-            last_instruction and
-            isinstance(last_instruction, MagneticTransfer) and
-            last_instruction.data.get("magnetic_head") == head
+                new_instruction is False and
+                last_instruction and
+                isinstance(last_instruction, MagneticTransfer) and
+                last_instruction.data.get("magnetic_head") == head
```
joshhacksthings, line 5604: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -6341,7 +6340,7 @@ def spectrophotometry(self, dataref, obj, groups, interval=None,
     def transfer(self, source, destination, volume, rows=1, columns=1,
                  source_liquid=LiquidClass,
                  destination_liquid=LiquidClass,
-                 method=Transfer, one_tip=False):
+                 method=Transfer, one_tip=False, density=None, mode=None):
```
joshhacksthings, line 6343: see above on refactoring

autoprotocol/protocol.py
```diff
@@ -6521,6 +6527,19 @@ def location_helper(source, destination, volume, method):
...
+                density = [density] * count
+            density = [parse_unit(d, "mg/ml") for d in density]
+            for d in density:
+                if d.magnitude <= 0:
+                    raise ValueError(
+                        "Density: {} must be a value larger than 0."
+                        "".format(d)
+                    )
+        else:
+            density = [density] * count
```
joshhacksthings, line 6541: if the `if density` above (line 6530) evaluates to False, what happens here?

autoprotocol/protocol.py
```diff
@@ -6593,20 +6612,24 @@ def location_helper(source, destination, volume, method):
...
             while remaining_vol > Unit(0, "ul"):
                 transfer_vol = min(remaining_vol, max_tip_capacity)
                 if one_tip is True:
-                    locations += location_helper(src, des, transfer_vol, met)
+                    locations += location_helper(src, des, transfer_vol, met, dens)
                 else:
+                    location_transports = location_helper(src, des, transfer_vol, met, dens)
+                    source_transports = location_transports[0]["transports"]
+                    instruction_mode = mode
+                    if instruction_mode is None:
```
joshhacksthings, line 6626: this can be `if not instruction_mode`

autoprotocol/protocol.py
```diff
@@ -6618,11 +6641,14 @@ def location_helper(source, destination, volume, method):
 
         # if one tip is true then there's a locations list
         if locations:
+            source_transports = locations[0]["transports"]
+            if mode is None:
```
joshhacksthings, line 6645: this should be `if not mode:`

autoprotocol/protocol.py
```diff
@@ -6618,11 +6641,14 @@ def location_helper(source, destination, volume, method):
...
+                mode = LiquidHandle.builders.mode(source_transports, mode)
```
joshhacksthings, line 6646: I don't know how useful calling `LiquidHandle.builders.mode(source_transports, None)` would be, if `mode` is not provided. I'd argue that `mode` argument for this builder is unnecessary

autoprotocol/protocol.py
```diff
@@ -6734,6 +6762,7 @@ def mix(self, well, volume, rows=1, columns=1,
         --------
         Mix : base LiquidHandleMethod for mix operations
         """
+
```
joshhacksthings, line 6765: i'd prefer not to reformat here, if this code is untouched

autoprotocol/protocol.py
```diff
@@ -6842,11 +6871,16 @@ def location_helper(aliquot, volume, method):
             if one_tip is True:
                 locations += location_helper(wel, vol, met)
             else:
+                location_transports = location_helper(wel, vol, met)
+                source_transports = location_transports[0]["transports"]
+                instruction_mode = mode
+                if instruction_mode is None:
```
joshhacksthings, line 6877: this can be `if not instruction_mode`

autoprotocol/protocol.py
```diff
@@ -6842,11 +6871,16 @@ def location_helper(aliquot, volume, method):
             if one_tip is True:
                 locations += location_helper(wel, vol, met)
             else:
+                location_transports = location_helper(wel, vol, met)
+                source_transports = location_transports[0]["transports"]
+                instruction_mode = mode
```
joshhacksthings, line 6876: I'm confused why you need to assign mode to a new object named `instruction_mode`. Can we just leave this as `mode` and refer to it as such below?

autoprotocol/protocol.py
```diff
@@ -6857,11 +6891,14 @@ def location_helper(aliquot, volume, method):
 
         # if one tip is true then there's a locations list
         if locations:
+            source_transports = locations[0]["transports"]
+            if mode is None:
```
joshhacksthings, line 6895: `if not mode:`

autoprotocol/protocol.py
```diff
@@ -6911,6 +6948,7 @@ def spread(self, source, dest, volume="50:microliter",
         TypeError
             If specified destination is not of type Well
         """
+
```
joshhacksthings, line 6951: do we need to reformat?

test/liquid_handle_unit_test.py
```diff
@@ -124,6 +124,7 @@ class LiquidHandleMethodTester(object):
     )
     asp_transport = LiquidHandle.builders.transport(
         volume=Unit(1, "uL"),
+        density=None,
```
joshhacksthings, line 127: We shouldn't be changing old tests to meet new specs, this means this isn't backwards compatible. My comment above about moving the density argument to the end of the function argument list should fix this (and you should then be able to remove `density=None` to get this test to pass again)

autoprotocol/builders.py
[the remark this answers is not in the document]
dzleidig: Ah, missed that, all good.

autoprotocol/builders.py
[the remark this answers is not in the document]
aota001: mode should default to `air_displacement` for `air` `default` or `none` liquid class types; and `positive_displacement` for `viscous` liquid class type. Currently, we don't specify what mode to use, because `air_displacement` is being hard-coded to be used. However, a user should be able to specify `mode` if needed, and if not specified, we should be able to resolve `mode` to whatever it fits... e.g. if a user wants to transfer glycerol, it should default to using `positive_displacement`, unless specified otherwise. This `mode` method allows for automatic resolution of mode depending on the liquid_class. Also, if say, glycerol is 'viscous' type, but if diluted enough, can be transferred using `air_displacement`. Hence, this will kick in only `if mode is None`. If user has a `mode` specified, that should be respected. Also, if (by any chance) there are mutiple `mode` specified in one liquid handle, that should be flagged... The behavior could also depend on how we add such aliquot propeties(?), tip types (since air-displacement and positive-displacement will use different tip types), etc, in the future as well....

autoprotocol/builders.py
aota001: As mentioned earlier, user should have flexibility to choose what dispense mode to use if needed. Liquid handle should be smart enough to resolve based on the liquid class of the sample, but depending on how viscous the sample is (e.g. depending on the concentration of solvent, etc), user should be able to pick whatever `mode` they want.

aota001: Having that said, I don't mind using dictionary for more legible code...

dzleidig: Sure, I think having a default mode mapping/dictionary would definitely make the code more legible. This is a bit tricky to walk through.

autoprotocol/builders.py
aota001: Hmm... good point to both... Since this is a heavily used functions, (and also nested with bunch of subfunctions), I'm leaning toward keeping it as a List for now as well...

autoprotocol/builders.py
```diff
@@ -1305,7 +1309,7 @@ def location(self, location=None, transports=None):
             "transports": transports
         }
 
-    def transport(self, volume=None, pump_override_volume=None,
+    def transport(self, volume=None, density=None, pump_override_volume=None,
```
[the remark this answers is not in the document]
aota001: Ah. good point. I'll fix that.

autoprotocol/builders.py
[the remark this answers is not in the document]
dzleidig: So if they specify a mode here, we want to check that this mode does not clash with their liquid types?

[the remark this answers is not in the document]
dzleidig: Maybe `desired_mode` would be more descriptive.

autoprotocol/builders.py
```diff
@@ -653,6 +654,7 @@ def shake_after(self, duration, frequency=None, path=None, amplitude=None):
 
 
 class SpectrophotometryBuilders(InstructionBuilders):
+    # pylint: disable=C0103
```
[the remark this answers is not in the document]
aota001: I included this one b/c it was failing on tox, and made this change before I realized that all R/C codes are ignored. lol. Will revert.

autoprotocol/builders.py
```diff
@@ -1184,7 +1186,8 @@ def position_z_calculated(self, wells, heuristic=None):
         heuristic: str, optional
             Must be one of "max_mean_read_without_saturation" or
             "closest_distance_without_saturation".
-            Please refer to `ASC-041 <http://autoprotocol.org/ascs/#ASC-040>`_ for the full explanation
```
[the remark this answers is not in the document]
aota001: Not sure why the change here... reverting.

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            if mode not in self.dispense_modes:
+                raise ValueError(
+                    "mode: {} must be one of the valid modes: {}"
+                    "".format(mode, self.dispense_modes)
+                )
+        modes = []
+        # get mode for each liquid_class if not specified already from the
+        # user input.
+        for liquid in non_air_classes:
+            if liquid is not None:
```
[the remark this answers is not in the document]
aota001: 👍

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
             "flowrate": flowrate
         }
 
+    def mode(self, transports=None, mode=None):
+        """Helper for selecting dispense mode based on liquid_class name
```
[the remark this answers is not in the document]
aota001: will do!

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                    raise TypeError(
+                        "liquid: {} is not of type str or None".format(liquid)
+                    )
+                if liquid not in self.liquid_classes:
+                    raise ValueError(
+                        "liquid_class: {} must be one of the valid classes: {} "
+                        "".format(liquid, self.liquid_classes)
+                    )
+                if mode is None:
+                    if liquid == "viscous":
```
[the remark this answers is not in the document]
aota001: will do!

autoprotocol/builders.py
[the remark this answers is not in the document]
aota001: this was giving lint error. 'line too long'

autoprotocol/builders.py
aota001: Hmmm. liquid class is a list, but this parameter takes in a value from the list, instead of taking in the whole list as an argument, though?

autoprotocol/builders.py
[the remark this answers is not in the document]
aota001: Right. I'll rename the method, and will add more documentation as Josh requested.

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            liquid is not of str
+        ValueError
+            liquid_class is not a valid value
+        ValueError
+            multiple liquid_class exists in one LiquidHandle
+        """
+        liquid_classes = set([transport["mode_params"]["liquid_class"]
+                              for transport in transports])
+        # remove automatically added 'air' class from blowout, etc.
+        non_air_classes = [liq for liq in liquid_classes if liq != "air"]
```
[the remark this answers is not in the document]
aota001: If the liquid_class happens to only contain "air", it will default to "air_displacement" :)

autoprotocol/builders.py
```diff
@@ -1738,6 +1747,81 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                    "".format(mode, self.dispense_modes)
+                )
+        modes = []
+        # get mode for each liquid_class if not specified already from the
+        # user input.
+        for liquid in non_air_classes:
+            if liquid is not None:
+                if not isinstance(liquid, str):
+                    raise TypeError(
+                        "liquid: {} is not of type str or None".format(liquid)
```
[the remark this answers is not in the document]
aota001: lol. I agree. Updated.

autoprotocol/builders.py
aota001: Yes, that is the intent here.

autoprotocol/builders.py
aota001: Ah. Will update the message.

autoprotocol/builders.py
aota001: This will pick on cases where modes are all "air", all None, or mixture of the two. Updating the comment.

autoprotocol/builders.py
aota001: I see. I'll update.

autoprotocol/builders.py
aota001: Actually, I take it back. Moving this earlier. :)

autoprotocol/builders.py
aota001: same as above... didn't realize tox ignores R/C codes. reverting!

autoprotocol/builders.py
aota001: ya... this was another lint error thing. reverting.

autoprotocol/protocol.py
aota001: this is intended

autoprotocol/protocol.py
aota001: Whether density is specified or is None, density will be turned into a list of [density] (line 6540-6541). The `if` statement here checks `density` value is valid if provided. 
... it is not checking for the list length if density is provided in a list, so I'm updating with additional test.

autoprotocol/protocol.py
aota001: so true. removed `if mode is None:`

autoprotocol/protocol.py
aota001: It is there because this happens inside a `for` loop. i.e. if I just call it `mode` it will overwrite the input `mode` the second time it goes through loop.

test/liquid_handle_unit_test.py
aota001: removed. passes fine.

aota001:
I think I addressed all the comments. @joshhacksthings @dzleidig If you would review one more time, that would be appreciated! Thanks!

autoprotocol/builders.py
aota001: Actually, changed it to `str` since that's what the value type of liquid_class should be.

autoprotocol/builders.py
```diff
@@ -1738,6 +1745,154 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            if len(set(modes)) > 1:
+                raise ValueError(
+                    "There are multiple liquid classes which could potentially"
+                    "have different modes: {}."
+                    "Please specify the mode to be used from: {}. Only one mode"
+                    "is allowed per transfer."
+                    "".format(set(modes), self.dispense_modes)
+                )
+
+            mode = list(modes)[0]
```
dzleidig, line 1892: Why are we recasting `modes` to a list? can't we just do` modes[0]`?

autoprotocol/builders.py
```diff
@@ -1738,6 +1745,154 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            # if liquid classes only contain "air" type, mode is returned.
+            if not non_air_classes:
+                return liquid_class_to_dispense_modes["air"]
+
+            modes = []
+            # get mode for each liquid_class if not specified already from the
+            # user input.
+            for liquid in non_air_classes:
+                # resolve mode for different liquid classes (except for None).
+                if liquid:
```
dzleidig, line 1857: Can we filter out the `None`s when populating `non_air_classes` to begin with? Just trying to reduce nesting here, this is still a bit involved.

autoprotocol/builders.py
```diff
@@ -1738,6 +1745,154 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                    "".format(mode, self.dispense_modes)
+                )
+        else:
+            # get liquid_classes from transport input
+            liquid_classes = set([transport["mode_params"]["liquid_class"]
+                                  for transport in transports])
+            # remove automatically added 'air' class from blowout, etc.
+            non_air_classes = [liq for liq in liquid_classes if liq != "air"]
+            # if liquid classes only contain "air" type, mode is returned.
+            if not non_air_classes:
```
dzleidig, line 1849: Double-negative here ("not non-...") followed by a mid-method return. Can we try to improve this without adding additional nesting? Creating a new method is fine.

autoprotocol/builders.py
```diff
@@ -1738,6 +1745,154 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                            "is updated.".format(
+                                set(modes), liquid_class_to_dispense_modes
+                            )
+                        )
+                    modes.append(liquid_class_to_dispense_modes[liquid])
+
+            # if liquid class only contains "air" and/or None, default to
+            # "air_displacement".
+            if not modes:
+                modes.append("air_displacement")
```
dzleidig, line 1880: Probably should use the dictionary here `liquid_class_to_dispense_modes["air"].`

autoprotocol/builders.py
aota001: modes is a `set`, so I'm turning it into a `list` to take the first element. Not sure if there is an easier way...?

autoprotocol/builders.py
aota001: Ya... I contemplated about that, too... Current behavior is that `None` gets `air_displacement` mode by default, so it would not be an issue to filter out both `air` and `None`, but if that behavior changes in the future, then we'll have to worry about having transports with just `air` alone, `None` alone and a mixture of `air` and `None`. I'm passing through all the `None`, and assigning `air_displacement` if the rest of the classes are all `None` explicitly. (`air` can be filtered out at the beginning since `air` can be in the present with any other liquid class types.)

autoprotocol/builders.py
aota001: Hmm. Not the best way to state 'if there is no non-air classes' here... I think it is easier than checking for 'transports that only contain 'air' and nothing else' (and state 'if it only contains air'), since I'll have to check for what other liquid_class is present in a list of transports later... I can change variable name if 'non-air...' doesn't read right/weird? Also, I think it's easier to just return the value here since if transports only contain `air` no further logic is necessary. I can move this to a later point, but I'll probably need to squeeze in another nested `if` anyways, since if transports only contain `air`,`modes` should be treated differently...

autoprotocol/builders.py
aota001: oops. missed this one. updating!

autoprotocol/builders.py
dzleidig: modes is a List...`modes = []`

autoprotocol/builders.py
dzleidig: Ok, I see, I'm alright with this then.

autoprotocol/builders.py
dzleidig: I mean, is this statement even needed?

If `non_air_classes` is empty, the for loop will skip and we will go directly to 1886, which has the same effect as returning immediately with `liquid_class_to_dispense_modes["air"]`.

```
non_air_classes = [liq for liq in liquid_classes if liq != "air"]
// non_air_classes = []
 modes = []
 for liquid in non_air_classes: //skipped
    ...
if not modes: // True, line 1186
   modes.append(liquid_class_to_dispense_modes["air"])
```

autoprotocol/builders.py
aota001: ya... thanks! I think `liquid_handle` just requires some refactoring in general, and when we get around to it, we can improve this. :)

autoprotocol/builders.py
aota001: This `modes` will also be empty if everything else in the `transports` is `None`. I like to have different behavior for when it contains `air` vs `None`, so I'll still need to have another logic here to separate the two. I don't mind changing if it's easier to read, tho. I don't have a strong preference one way or another... I'll think about it a bit more to see if i can make this better. Meanwhile, let me know if you have suggestions or strong preference for how it reads!

dzleidig: Hmm, I'm not following. Can we meet to discuss?

dzleidig: You say different behavior, but it's the same behavior, right? The same effect...

autoprotocol/builders.py
aota001: my bad!! I think I was using `set` at one point, or got confused with liquid_classes. :p Updating!

autoprotocol/builders.py
aota001: Removed double-negative condition :)

autoprotocol/builders.py
```diff
@@ -1305,8 +1308,8 @@ def location(self, location=None, transports=None):
             "transports": transports
         }
 
-    def transport(self, volume=None, pump_override_volume=None,
-                  flowrate=None, delay_time=None, mode_params=None):
+    def transport(self, volume=None, pump_override_volume=None, flowrate=None,
```
joshhacksthings, line 1311: super duper picky nit: can you add the line break after `pump_override_volume` just to keep first line consistent with previous versions. Then git will only highlight the second line as being updated.

joshhacksthings: (in other words, move `flowrate=None` to the line below along with `delay_time`, etc.)

autoprotocol/protocol.py
```diff
@@ -5507,7 +5506,7 @@ def image(self, ref, mode, dataref, num_images=1, backlighting=None,
             raise TypeError("num_images must be a positive integer.")
         if magnification:
             if not isinstance(magnification, (float, int)) or \
-               magnification <= 0:
+                magnification <= 0:
```
joshhacksthings, line 5509: nit: don't refactor code out of scope of PR

autoprotocol/protocol.py
joshhacksthings, line 1536: nit: don't refactor code out of scope of PR

autoprotocol/protocol.py
```diff
@@ -1385,7 +1385,7 @@ def illuminaseq(self, flowcell, lanes, sequencer, mode, index, library_size,
                              " modes: {}.You specified: {}."
                              "".format(sequencer,
                                        ', '.join(valid_sequencers[
-                                                 sequencer]["modes"]),
+                                                sequencer]["modes"]),
```
joshhacksthings, line 1388: nit: don't refactor code out of scope of PR

autoprotocol/builders.py
```diff
@@ -1324,26 +1327,31 @@ def transport(self, volume=None, pump_override_volume=None,
...
+            Density of the liquid to be aspirated/dispensed in mg/ml.
 
 
         Returns
         -------
         dict
             transport parameters for a LiquidHandle instruction
         """
-        if volume is not None:
+        if volume:
```
rhysormond, line 1339: we actually want the stricter checks of `is not None` here since AFAIK `0:microliters` is falsy and therefore wouldn't hit this check

autoprotocol/builders.py
```diff
@@ -1738,6 +1746,146 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+            multiple liquid_class exists in one LiquidHandle
+        """
+        # dict for default dispense mode for each liquid_class
+        liquid_class_to_dispense_mode = {
+            "air": "air_displacement",
+            "default": "air_displacement",
+            "viscous": "positive_displacement"
+        }
+        if mode:
+            if not isinstance(mode, str):
```
rhysormond, line 1834: is this check redundant with the below one? there's no way for a non-string thing to be in dispense_modes

autoprotocol/builders.py
```diff
@@ -1738,6 +1746,146 @@ def blowout(self, volume, initial_z, flowrate=None):
...
+                    "is a new liquid_class, make sure dictionary: {}"
+                    "is updated.".format(
+                        other_class, liquid_class_to_dispense_mode
+                    )
+                )
+            modes.add(liquid_class_to_dispense_mode[other_class])
+
+        # return error if there are incompatible liquid_class a set of
+        # transports.
+        if len(set(modes)) > 1:
```
rhysormond, line 1878: is there any chance we have no modes in which case the below `modes[0]` will blow up?

autoprotocol/builders.py
aota001: kk

autoprotocol/protocol.py
aota001: undone

autoprotocol/builders.py
aota001: oh good to know... thanks!

autoprotocol/builders.py
dzleidig: There shouldn't be.

1851:
```
        if not other_classes:
            return liquid_class_to_dispense_mode["air"]
```

Otherwise it goes into the for loop, where unless an exception is thrown, a mode is added for each iteration.

autoprotocol/builders.py
aota001: I guess so? It just catches different error type. I can remove it.

autoprotocol/builders.py
aota001: Since `modes` can only get value by looking up dictionary, and there is an error thrown to catch such error, so there shouldn't be a case where `modes` will be empty at this point. I guess there used to be `if not modes: ...` for a different purpose before... I can add a check just in case.

aota001:
thanks @rhysormond. Let me know if it looks ok now. :)

PR #230, opened 2020-04-15
changing how Protocol.propagate_properties copies properties to destination wells

EribertoLopez:
https://strateos.atlassian.net/jira/software/projects/STDEV/boards/58?selectedIssue=STDEV-35 

User warnings are preventing protocol package from building and uploading to web. This is due to setting Protocol.propagate_properties to True, because it will use the .add_properties method instead of the .set_properties method. The .add_properties method purposefully uses user warnings when aliquot properties are overwritten. By changing to .set_properties, we can avoid building failures and copy the aliquot properties to the destination aliquot.

### DIFF
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1827,42 +1827,45 @@ class LiquidHandleBuilders(InstructionBuilders):
         ValueError
             multiple liquid_class exists in one LiquidHandle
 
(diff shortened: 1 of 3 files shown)


EribertoLopez:
> rebase, then you're good!

rebased, and up to date

PR #234, opened 2020-05-02
feat: upadate README with Dev Env setup

aota001:
Adding Dev Env section in README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,64 +41,14 @@ Dev Env Setup
 -------------
 
 Installation:
+    Current Python version supported for development is :code:`v3.5.6`.
+
     python_
 
-Python Version:
-    Current version accepted for development is :code:`v3.5.6`.
-    See python_
-
-Install Multiple Python versions (Optional):
-    There are several ways to install multiple Python versions if needed.
-
-    May use pyenv_ to install multiple Python versions, specify global, local or
-    temporary Python version, and to switch between different versions.
-
-    To install different versions of Python, and to list installed pyenv
-    versions:
-
-    .. code-block:: bash
-
-        pyenv install <python_version>
-        pyenv versions
-
-    To activate pyenv:
-
-    .. code-block:: bash
-
-        pyenv activate <python_version>
-
-    To set global, local or shell-specific Python version:
-
-    .. code-block:: bash
-
-        pyenv global <python_version>
-        pyenv local <python_version>
-        pyenv shell <python_version>
-
-    To allow auto-activation of pyenv:
-
-    .. code-block:: bash
-
-        echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n
-        eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
-
-    Virtual Environment:
-        virtualenv is a plugin for pyenv that manages virtual environments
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -37,6 +37,72 @@ Alternatively to get more up-to-date features:
...
 
+Dev Env Setup
+-------------
+
+Installation:
+    Python https://www.python.org/
+
+    Pyenv https://github.com/pyenv/pyenv#installation
+
+Python Version:
```
rhysormond, line 48: hmmm rather than encoding this in the readme what do we think about just adding a `.python-version` file to enforce this a bit more specifically

README.rst
```diff
@@ -124,3 +190,4 @@ For more information, see CONTRIBUTING_.
 .. _changelog: http://autoprotocol-python.readthedocs.io/en/latest/changelog.html
 .. _CONTRIBUTING: http://autoprotocol-python.readthedocs.io/en/latest/CONTRIBUTING.html
 .. _documentation: http://autoprotocol-python.readthedocs.org/en/latest/
+.. _devenvsetup: https://strateos.atlassian.net/wiki/spaces/ENG/pages/191660547/Python+environment+setup
```
rhysormond, line 193: I'm not super keen on linking our own internal documentation in a public repository

README.rst
```diff
@@ -37,6 +37,72 @@ Alternatively to get more up-to-date features:
...
+
+    To install different versions of Python, and to list installed pyenv
+    versions:
+
+    .. code-block:: bash
+
+        pyenv install <python_version>
+        pyenv versions
+
+    To activate pyenv:
```
rhysormond, line 63: While we use pyenv internally I'm not sure how I feel about pushing it along with APPY. Other folks may very well be using pipenv, conda (very likely among APPY users), or some other alternative.

README.rst
aota001: Ah. I can change this.

README.rst
aota001: Ya, @yangchoo mentioned the same. I forgot this is a public repo. Will remove it.

README.rst
aota001: hmm... I guess I did initially install conda before I was told to use pyenv. lol. Will bring this up so I can get other opinions. Otherwise, we can mark python version as requirement, and have pyenv, etc, as optional/suggested setup to leave it up to the developer?

aota001:
@rhysormond  After reading some other python libraries to see what information they provide, I removed whole section on `pyenv` specific setup instructions, and I left it with supported python version, and a remark 'virtual environment is highly recommended' for dev, but leaving developer to use whichever tool. Think this is more appropriate?

README.rst
```diff
@@ -37,6 +37,19 @@ Alternatively to get more up-to-date features:
 
 check the changelog_ for information about features included on `master` but not yet released.
 
+Dev Env Setup
+-------------
+
+Installation:
+    Current Python version supported for development is :code:`v3.5.6`.
```
yangchoo, line 44: nit: Let's say refer to `setup.py`, `python_requires` line to get the minimum supported Python version. That way we don't have to remember to update two places

PR #235, opened 2020-05-11
feat: support Protocol subclasses for harness.run

yangchoo:
We currently default to instantiating a Protocol object in harness.run
However, that precludes the user from creating their own subclasses of
Protocol and using that in the harness.run method.

This would enable that use case, which would increase the flexibility
of the manifest/harness system.


yangchoo:
@EribertoLopez thanks! Would you mind re-reviewing, had to rebase with latest master, thanks :)

PR #241, opened 2020-05-20
build: Upgrade to Python 36

yangchoo:
Upgrade base Python version to Python 3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ env:
 sudo: false
 language: python
 python:
-- '3.5'
 - '3.6'
 - '3.7'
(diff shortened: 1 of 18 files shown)


yangchoo:
> do we need to express anything else to denote this as a breaking change? e.g. what implications does this have in terms of semantic versioning. i assume with the setup.py changes (specifically the python_requires statement) that things should "just work" in terms of packaging and ensuring that folks on 3.5 don't install this from pypi?

Nope! `python_requires` neatly handles this, so there isn't anything else we need to do. Basically, if you're on 3.5, the previous version would be the final release

PR #243, opened 2020-05-21
build: pre-commit infrastructure and update travis

yangchoo:
Add pre-commit infrastructure to help with basic lint tasks. This is in
preparation of adding an autoformatter for the repo.

As part of this, the travis file is reorganized to help with clearer
delineation of jobs.

Documention is also updated, with the Contributing section updated with
more explicit testing steps and the new environment and lint setup.

This includes all the fixes from the pre-commit hooks added.


yangchoo:
Please do check out the Travis build for the new test matrix: https://travis-ci.org/github/autoprotocol/autoprotocol-python/builds/689570208

.travis.yml
```diff
@@ -5,12 +5,25 @@ env:
...
-install: pip install tox-travis
-script: tox
+jobs:
+  include:
+  - python: '3.6'
+    env: TOXENV=docs
+  - python: '3.6'
+    env: TOXENV=lint
+  - python: '3.6'
+    env: TOXENV=clean,py36,stats
```
rhysormond, line 15: is the clean step here necessary still?

CONTRIBUTING.rst
```diff
@@ -11,6 +11,58 @@ Features and Bugs
...
+Python Virtual Environment (Optional):
+    Use of virtual environment to isolate development environment is
+    highly recommended. There are several tools available such as
+    conda_ and pyenv_.
+
+Dependencies:
+    We recommend first activating your desired virtualenv, then
+    installing the dependencies using the snippet below.
+
+.. code-block:: bash
```
rhysormond, line 29: extremely minor nit: these aren't really using anything bash-specific and could just be `sh` which *appears* to be a supported tag

rhysormond:
somewhat related, can we configure travis so that it's using something newer than xenial?

.travis.yml
yangchoo: ah, yea, its to blow away coverage artifacts if present. It seems common in other travis integration libraries, so I assume its doing something but haven't validated

CONTRIBUTING.rst
yangchoo: cool! will do

yangchoo: yup sh works

yangchoo:
yea, could use Bionic (https://docs.travis-ci.com/user/reference/linux/).

It shouldn't really matter since we're using mostly pip/tox for the running, so it should be fine to upgrade. I'll try it

PR #245, opened 2020-05-21
Add expected propagate_properties behavior tests

EribertoLopez:
Adding test for expected propagate property behavior when the destination well has prior properties.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,8 @@ Changelog
 =========
 
 * :support:`247` Add `black` as auto-formatter to pre-commit workflow
-* :bug:`244` Revert Protocol.propagate_properties to use Well.add_properties
+* :support:`245` Add expected propagate_properties behavior tests
+* :bug:`244 major` Revert Protocol.propagate_properties to use Well.add_properties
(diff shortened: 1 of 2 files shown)


test/protocol_test.py
```diff
@@ -2762,6 +2762,21 @@ def test_can_transfer_properties(self):
             self.container.well(0).properties
         )
 
+    def test_can_append_properties(self):
+        """Expected behavior when propagating properties to wells with prior properties."""
+        self.p.propagate_properties = True
+        self.container.well(0).set_properties({"foo": ["bar0"], "bar": "foo1"})
+        self.container.well(1).set_properties({"foo": ["bar1"], "bar":"foo2"})
```
yangchoo, line 2769: nit: whitespace, `"bar":"foo2"` -> `"bar": "foo2"`

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :support:`245` Add expected propagate_properties behavior tests
 * :bug:`244` Revert Protocol.propagate_properties to use Well.add_properties
```
yangchoo, line 6: something to note here is that we probably want a `major` flag attached to `244` unless we want an intermediate bugfix release
i..e :bug:`244 major`

PR #246, opened 2020-05-22
Added missing test cases for existing provision instruction

ramu-phanimukla:
This PR is addressing below:

- Also fixed an issue related to provisioning multiple wells
- Adding missed test cases


dzleidig:
Looks like there are some build issues, @ramu-phanimukla . Accessing the test data directory. See status checks.

https://travis-ci.org/github/autoprotocol/autoprotocol-python/builds/689859596

dzleidig:
Looks like it can be reproduced locally. When running pytest via the IDE it works but via ` python setup.py test` it does not work. Something up with the relative and working directory....

dzleidig:
Yeah ok, so when it is run in CI, the working directory is actually one up.

```
import json
import os


class TestUtils:

    @staticmethod
    def read_json_file(file_path: str):
        print(f"currrent working directory: {os.getcwd()}")
        file = open('./test/data/{0}'.format(file_path))
        data = json.load(file)
        return json.dumps(data, indent=2, sort_keys=True)
```


`currrent working directory: /Users/dylanleidig/dev/autoprotocol/autoprotocol-python`

So, this works: `file = open('./test/data/{0}'.format(file_path))`.

Not sure the best way to resolve this so it works regardless of working directory. Maybe have the call pass in the working directory too.

test/test_util.py
```diff
@@ -0,0 +1,10 @@
+import json
+
+
+class TestUtils:
+
+    @staticmethod
+    def read_json_file(file_path: str):
+        file = open('./data/{0}'.format(file_path))
```
yangchoo, line 8: as @dzleidig mentioned in his comment, this should be relative to the root directory

test/provision_test.py
```diff
@@ -0,0 +1,108 @@
...
+            self.p.provision("rs17gmh5wafm5p", self.w1, "500:microliter")
+
+    # TODO - If all container-types defined in `container_type.py` are of type Well, then code lines
+    # TODO - between 4960-4962 in protocol.py doesn't make any sense
+    # def test_attempt_to_provision_with_unsupported_container_type(self):
+    #     p = Protocol()
+    #     w1 = p.ref("w1", None, cont_type="micro-1.5", discard=True) \
+    #         .well(0).set_volume("2:microliter")
+    #     with pytest.raises(TypeError):
+    #         p.provision("rs17gmh5wafm5p", w1, "50:microliter")
```
yangchoo, line 32: nit: remember to remove this before merging. Generally will prefer to surface these as issues rather than ephemerally in code, since the line numbers/context may change

test/provision_test.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_with_destinations_count_not_same_as_volumes(self):
+        volumes = ["50:microliter", "20:microliter"]
+        with pytest.raises(RuntimeError):
+            self.p.provision("rs17gmh5wafm5p", self.w1, volumes)
+
+    # TODO - This test should pass as there is no definition of `ml` in http://autoprotocol.org/specification/.
+    # TODO - Since the current code is able to translate `ml` to `microliter`, this test case is failing
+    # def test_with_invalid_unit(self):
+    #     with pytest.raises(TypeError):
+    #         self.p.provision("rs17gmh5wafm5p", self.w1, "20:ml")
```
yangchoo, line 47: ah. this is expected behavior currently. The Units library provided in ApPy is more powerful/less strict than the Autoprotocol spec, i.e. it has deserialization support for short-hand such as `ml`.
Ultimately, this still gets serialized into `milliliter` and not `ml`, so I think that's fine to provide that convenience to the writer. 
It'll become a concern if we do end up serializing to an unsupported unit.

yangchoo: Broadly though, we've talked about potentially expanding the units in the spec to a bigger range. That's a separate topic which has some lively discussion, though we haven't seen real world needs for that yet.

test/provision_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+class TestProvision(object):
+    p = Protocol()
+    w1 = p.ref("w1", None, cont_type="96-pcr", discard=True) \
+        .well(0).set_volume("2:microliter")
+
+    def test_provision_well_capacity(self):
+        self.p.provision("rs17gmh5wafm5p", self.w1, "50:microliter")
+        actual_protocol_as_json = json.dumps(self.p.as_dict(), indent=2, sort_keys=True)
+        expected_protocol_as_json = TestUtils.read_json_file('expected_provision.json')
```
yangchoo, line 17: hmm I'm wary of this pattern. Technically, we're only modifying the provision code, so we don't want these tests to break if we were to make a separate change to a top-level key for example which is independent of the provision code.

Its debatable whether we want to use json.dumps directly, since that adds an additional fixed dependency for the serialization check. Currently we mostly check against the generated Instruction data-structure, since that's what the method directly does, and assume that the serialization portion is decoupled and works.

autoprotocol/protocol.py
```diff
@@ -4979,8 +4979,9 @@ def provision(self, resource_id, dests, volumes):
                     d.container):
                 self.instructions[-1].to.append(xfer)
             else:
-                return self._append_and_return(Provision(resource_id,
+                self._append_and_return(Provision(resource_id,
                                                          dest_group))
+        return self
```
yangchoo, line 4984: didn't spot this, but generally we return the generated Instruction here, and not the protocol

ramu-phanimukla: Actually there is no need to return as the method invocation is on the same Protocol object. I have validated it and works fine.

test/provision_test.py
ramu-phanimukla: Although we made changes to provision code alone, it is referring to the object `refs` in the protocol object. So, we should test provision in conjunction with `refs`. Since `refs` is a child of protocol object, I am using complete object for comparison. Moreover if we have the flexibility to create just `Provision` object alone, then I would have written test cases just around that. Since provision can be created via protocol, we should test against the entirety of protocol object.

The main intent of using json usage is to avoid boiler-plate code for assertions. There are no custom serialization in any of these methods.

test/provision_test.py
ramu-phanimukla: As a specification I am implementing according to what is listed at http://autoprotocol.org/specification/. Here there is no mention of `ml`. We must call it out. Let's down the line if I am implementing this specification in say Java, I might implement to support only `meter, millimeter, micrometer`.

test/test_util.py
ramu-phanimukla: Fixed

test/provision_test.py
ramu-phanimukla: Intent is to bring the attention of core authors. Removed this code. Ideally I want to remove below code as well:
```
        if not is_valid_well(dests):
            raise TypeError("Destinations (dests) must be of type Well, "
                            "list of Wells, or WellGroup.")
```

ramu-phanimukla:
> Main thing here is figuring out the test pattern. As mentioned in one of the comments, would prefer if we encapsulate the tests here more, to test the `provision` method specifically rather than against an entire protocol json.
> 
> As for whether to test the `provision` json, I think that's something worth discussing. My preference is that we encapsulate the serialization behavior from the protocol data-structure insertion behavior, and have a separate Instruction serialization test.
> But that does introduce additional boilerplate, so its something to consider.

As a platform, if we are allowing someone to just submit a run containing only _provision_ instructions, then the pattern adopted is valid.

autoprotocol/protocol.py
```diff
@@ -4979,7 +4973,7 @@ def provision(self, resource_id, dests, volumes):
                     d.container):
                 self.instructions[-1].to.append(xfer)
             else:
-                return self._append_and_return(Provision(resource_id,
+                self._append_and_return(Provision(resource_id,
```
yangchoo, line 4976: hmm.. was going to write we should keep this, but I can see how this introduced a bug.

Basically the return here could potentially short-circuit the for-loop, which will fail in the case there are multiple wells.

This seems similar to the `transfer` case, where it could generate `n` LiquidHandle Instructions. Let's return a list here, similar to how its done there.

I.e. accumulate to some list here in this statement and return that list at the end. Let's change the type signature of the return above as well to `list(Provision)`

autoprotocol/protocol.py
```diff
@@ -4920,12 +4920,6 @@ def provision(self, resource_id, dests, volumes):
         ValueError
             Volume to provision exceeds max capacity of well
 
-        Returns
-        -------
-        Provision
-            Returns the :py:class:`autoprotocol.instruction.Provision`
-            instruction created from the specified parameters
-
```
yangchoo, line 4928: hmm.. I think we still want to ensure we return for this Instruction, that is something we do for all Instruction-creation methods. The intent there is that the user can get a handle on the Instruction and eventually feed it into a time constraint or similar construct.

I think pylint caught a bug here where it inspected the output and realized that we were not returning in all cases.

autoprotocol/protocol.py
```diff
@@ -4979,7 +4973,7 @@ def provision(self, resource_id, dests, volumes):
                     d.container):
                 self.instructions[-1].to.append(xfer)
```
yangchoo, line 4974: I think the issue with the lint warning is on this branch.. here, we're just appending without returning.

We probably want to return the modified instruction here (i.e. self.instructions[-1]). Generally the modify in-place isn't ideal.. but I think returning what we modified would make sense

test/provision_test.py
yangchoo: had a discussion offline about this. 
In summary, this is a viable approach if we were to just test against Instructions. It forces the developer to update each json file if a new Instruction field is added, which trades off test maintenance with potentially edge-case discovery. The hope there is that by forcing the developer to update each file, he may discover some dependent relationship. Its worth trialing this for this Instruction.

test/provision_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+class TestProvision(object):
+    p = Protocol()
+    w1 = p.ref("w1", None, cont_type="96-pcr", discard=True) \
+        .well(0).set_volume("2:microliter")
+
+    def test_provision_well_capacity(self):
+        self.p.provision("rs17gmh5wafm5p", self.w1, "50:microliter")
+        actual_protocol_as_json = json.dumps(self.p.as_dict()['instructions'], indent=2, sort_keys=True)
+        expected_protocol_as_json = TestUtils.read_json_file('expected_provision.json')
```
yangchoo, line 15: nit: let's update `protocol` -> `instruction` for subsequent references. i.e. `actual_protocol_as_json` -> `actual_instruction_as_json`, `expected_protocol_as_json` -> `expected_instruction_as_json`

autoprotocol/protocol.py
ramu-phanimukla: Returning `list(Provision)` also doesn't makes sense since we are adding additional instructions(`uncover` & `unseal`), conditionally. Anyhow I made changes to capture & return only `Provision` instructions to retrofit other existing methods implementation pattern. 

Generation observation: I see that `Protocol` class is following builder design pattern(https://refactoring.guru/design-patterns/builder/python/example) , where it is enhancing the protocol object. But at the same time, we are expecting each method to return only what is being added(which is more of like a utility method where we pass certain params and it returns some object)

autoprotocol/protocol.py
ramu-phanimukla: Made changes to return only `Provision` instructions.

BTW, in the previous commit I removed return type even from documentation. Not sure how my local `pre-commit` hook missed it

autoprotocol/protocol.py
ramu-phanimukla: Now that we are returning provision object, I think this comment is addressed

test/provision_test.py
ramu-phanimukla: Renamed

autoprotocol/protocol.py
```diff
@@ -5068,8 +5068,7 @@ def provision(self, resource_id, dests, volumes):
         Returns
         -------
         Provision
```
yangchoo, line 5070: this should probably be of type `list(Provision)`

autoprotocol/protocol.py
yangchoo: yup! let's update the return statement, but this looks good otherwise

test/provision_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class TestProvision(object):
+    p = Protocol()
+    w1 = (
+        p.ref("w1", None, cont_type="96-pcr", discard=True)
+        .well(0)
+        .set_volume("2:microliter")
+    )
+
```
joshhacksthings, line 15: you should make this a pytest fixture as your tests _could_ break, if run out of order. Additionally, the `Protocol` object is mutable with side-effecting methods. This will also remove redundant boilerplate if you declare your protocol, container, and well objects upfront, instead of declaring them every time.

test/provision_test.py
```diff
@@ -0,0 +1,185 @@
...
+            p.ref("w2", None, cont_type="96-pcr", discard=True)
+            .well(0)
+            .set_volume("2:microliter")
+        )
+        w3 = (
+            p.ref("w3", None, cont_type="96-pcr", discard=True)
+            .well(0)
+            .set_volume("2:microliter")
+        )
+        wells = [w1, w2, w3]
```
joshhacksthings, line 61: you could rewrite this using the `.wells_from` method to get this list of wells

ramu-phanimukla:
@joshhacksthings I had multiple conversations with @yangchoo on this approach using json comparison. We looked at both pros and cons of the approach and what value this approach is going to bring. We agreed that this sometimes will require changes across many files but will help in uncovering new scenarios to test for.

yangchoo:
@joshhacksthings I think your code feedback suggestions are legitimate improvements. At this stage, perhaps another PR to fix those things up may be fine.

In terms of the test pattern though, as Ramu said, we had discussions about the json comparison approach.

Not sure how to link to a resolved conversation, but the tl;dr snippet from above is that Ramu and I had a conversation and the concluding reason for trialing this approach is
> It forces the developer to update each json file if a new Instruction field is added, which trades off test maintenance with potentially edge-case discovery. The hope there is that by forcing the developer to update each file, he may discover some dependent relationship. Its worth trialing this for this Instruction.

I think one thing to emphasize here is that the value from this alternate approach only really works if a developer takes his/her time to change the test files and not do it in some scripted manner.

I echo most of your concerns with regards to introducing boilerplate and do share concerns about scaling sustainability (especially in the original proposal to test the entire Protocol json). 
However, I think its worth trialing this approach out due to the potential benefits above.

To emphasize, I'm definitely hesitant to roll this approach out to the entire repo. Would be happy to revisit the conclusions of the approach, especially if we're thinking of rolling this out more.

PR #247, opened 2020-05-22
build: Use black as autoformatter

yangchoo:
Add black as a dependency to pre-commit so that it can be run locally
and as part of CI infrastructure.

This also checks in the auto-formatting changes from black.

Using default black configuration, e.g. max line-length is 88


yangchoo:
From a skim, most of the changes are related to 
- consistency on quotes (generally `'` -> `"`)
- adding trailing commas
- indentations

PR #250, opened 2020-05-22
 build: Bump test dependencies

yangchoo:
As we're preparing for a major release, this is an opportunity to
upgrade some of our base dependencies.

This bumps our pytest dependency by one major version. Notably py35
support was dropped, and there has been multiple deprecations which we
are not affected by. There's also a bunch of bugfixes/improvements.

Pylint has been pinned to 2.5.2, which is the same as what we're running
in pre-commit.

Tox has been bumped to 3.15, which brings notable speedups.

### DIFF
--- a/autoprotocol/instruction.py
+++ b/autoprotocol/instruction.py
@@ -822,14 +822,14 @@ class Fluorescence(Instruction):
                 "calculated_from_wells": []
             }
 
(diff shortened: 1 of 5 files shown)


rhysormond:
Looks like there're some new lint warnings to resolve before landing but at a quick glance they all seem sane.

PR #252, opened 2020-05-27
docs: Use readthedocs config file

yangchoo:
We had a separate doc/requirements.txt file that was previously used for
maintaing documentation dependencies for readthedocs.org

Switching to the readthedocs config file would allow us to remove that
duplicated file and use the right build environment for generating docs.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,18 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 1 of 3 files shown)


yangchoo:
Referencing: https://docs.readthedocs.io/en/stable/config-file/v2.html#formats

Docs are currently failing to build, pointed readthedocs at this branch and its building fine now. Will point back to master once landed.

See https://autoprotocol-python.readthedocs.io/en/latest/changelog.html (and notice that this entry `[Support] #252: Switch to using .readthedocs.yml for docs build config` is present)

PR #254, opened 2020-05-29
Initial commit for supporting mass provision

ramu-phanimukla:
This is initial commit. More test cases will be added. Before I move far, need quick feedback


autoprotocol/protocol.py
```diff
@@ -5080,51 +5081,62 @@ def provision(self, resource_id, dests, volumes):
...
-                raise TypeError("Volume must be a string or Unit.")
+            amounts = [Unit(v) for v in amounts]
+        unique_measure_modes = set()
+        for amount in amounts:
+            if not isinstance(amount, (str, Unit)):
+                raise TypeError("Volume/Mass must be a string or Unit.")
+            if amount.unit in VOLUMETRIC_UNITS:
+                unique_measure_modes.add("volume")
+            elif amount.unit in MASS_UNITS:
+                unique_measure_modes.add("mass")
```
yangchoo, line 5102: hmm we should probably explicitly fail here instead of silently fail? (i.e. else case)

yangchoo, line 5102: If I understand the intention behind this change, the intention here is to ensure that we generate units which are part of the Autoprotocol-spec.

If that's the intention, I think it'll make more sense to check this during the final serialization portion, i.e. in https://github.com/autoprotocol/autoprotocol-python/blob/master/autoprotocol/protocol.py#L5810 or perhaps upstream in the `as_dict` call.

That way, it'll be more applied to all Unit references, not just provision. 

If we were to do that, something else we need to think about is we could potentially automatically convert it to the smallest/largest supported unit. For example, someone specifying "1000:nanoliter" is technically valid when converted to "1:microliter", and they're equivalent.

autoprotocol/protocol.py
```diff
@@ -5080,51 +5081,62 @@ def provision(self, resource_id, dests, volumes):
...
+                            self.provision(resource_id, d, diff)
+                        )
+                        diff -= diff
+                    continue
+
+                if d.volume:
+                    d.volume += amount
+                else:
+                    d.set_volume(amount)
+            self._remove_cover(d.container, "provision")
```
yangchoo, line 5138: remove cover call should happen before we start provisioning

autoprotocol/protocol.py
```diff
@@ -5080,51 +5081,62 @@ def provision(self, resource_id, dests, volumes):
...
             xfer["well"] = d
-            xfer["volume"] = v
+            xfer[measurement_mode] = amount
 
-            if d.volume:
-                d.volume += v
-            else:
-                d.set_volume(v)
-            dest_group.append(xfer)
+            if measurement_mode == "volume":
```
yangchoo, line 5115: We briefly talked about the possibility of doing the max_vol check by converting with density... Probably more of a nice-to-have

test/provision_test.py
```diff
@@ -183,3 +183,20 @@ def test_for_multiple_dispenses_of_resource_in_containers_larger_than_900ml(self
...
+        )
+        p.provision("rs17gmh5wafm5p", w1, "50:ug")
+        actual_instruction_as_json = json.dumps(
+            p.as_dict()["instructions"], indent=2, sort_keys=True
+        )
+        expected_instruction_as_json = TestUtils.read_json_file(
+            "provision_for_mass.json"
+        )
+
+        assert expected_instruction_as_json == actual_instruction_as_json
```
yangchoo, line 202: lets try to add more failure cases

autoprotocol/constants.py
```diff
@@ -35,3 +35,9 @@
     (0.6243159222654036, -0.14191423819668741),
     (0.563207308766221, 0.5152506858187523),
 ]
+
+MEASUREMENT_MODES = ["mass", "volume"]
+
+VOLUMETRIC_UNITS = ["liter", "milliliter", "microliter"]
+
+MASS_UNITS = ["gram", "milligram", "microgram"]
```
yangchoo, line 43: See below, something to consider is to abstract this out more generically

autoprotocol/protocol.py
ramu-phanimukla: Actually below code will catch(I know message is not capturing that).
```
        if len(unique_measure_modes) != 1:
            raise ValueError("Received amounts with more than one measurement mode")
```

Anyhow I added else block.

autoprotocol/protocol.py
ramu-phanimukla: Good catch.

autoprotocol/protocol.py
ramu-phanimukla: Since there is no consensus yet on this, I will wait on product owner to conform on the requirement. May be I will reach out to apps team try to get their input. But I prefer to have PO take final call and put it in the issue.

test/provision_test.py
ramu-phanimukla: My commit message was :

> This is initial commit. More test cases will be added. Before I move far, need quick feedback

test/data/provision_for_mass.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "measurement_mode": "mass",
```
joshhacksthings, line 3: just curious why `"measurement_mode"` is sorted to the top here when set to `"mass"`, but not for `"volume"`, when `json.dumps(sort_key=True)` sorts by keys.

autoprotocol/constants.py
joshhacksthings, line 43: there are more volumetric and mass units than these. is it possible not to hardcode here?

autoprotocol/protocol.py
```diff
@@ -5058,19 +5058,20 @@ def provision(self, resource_id, dests, volumes):
         TypeError
             If resource_id is not a string.
         RuntimeError
-            If length of the list of volumes specified does not match the number
+            If length of the list of volumes/masses specified does not match the number
             of destination wells specified.
         TypeError
-            If volume is not specified as a string or Unit (or a list of either)
+            If volume/mass is not specified as a string or Unit (or a list of either)
```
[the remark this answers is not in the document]
ramu-phanimukla: Makes sense. Fixed it through-out this method.

test/data/provision_for_mass.json
ramu-phanimukla: In json file can have the keys in any order. But at load time(with the option of `sort_keys=True`), it will sort the keys. 

The main reason for specifying the option of `sort_keys=True` is that, when developer defines json(with keys in any order) and there is no library to support comparison of logical structure and data of expected v/s actual json, I am relying on string comparison(for which keys need to be sorted)

autoprotocol/constants.py
ramu-phanimukla: @joshhacksthings I know there are more volumetric and mass units. As an implementer AP specification, I read http://autoprotocol.org/specification/#units and assumed that, in the scientific world only mentioned units makes sense and hence scoped down to this list.

As per the ASC https://github.com/strateos/autoprotocol_standard_changes/blob/master/Proposed/asc052.md, in order to populate `measurement_mode` attribute, I need to refer to these constants. If you can point me to some library which can return some indication of measurement made based on unit, I will get rid of these constants. Till then, we may have to retain(I will add other possible units for volume and mass). Let me know.

ramu-phanimukla:
> I would prefer if we separated out the Unit check from this diff... I think its definitely something worth considering, but it'll be something that'll make more sense if its consistently applied.
Can you please elaborate bit more on how you are envisioning this?

> We should probably also decide on if we want to do the mass check via density conversion.
Based on the comment https://strateos.atlassian.net/browse/SEA-240?focusedCommentId=16025, from product we are not performing any checks now.

autoprotocol/protocol.py
ramu-phanimukla: Based on [comment](https://strateos.atlassian.net/browse/SEA-240?focusedCommentId=16025)  from product, we are not adding any checks or logic as part of this US implementation

ramu-phanimukla:
@yangchoo & @joshhacksthings 

I removed static definitions from constants file and using string functions to decide measurement mode. Have a look at the changes. The main intention of adding those constants was trying to adhere to whatever was listed in [AP specification](http://autoprotocol.org/specification/#units).

autoprotocol/constants.py
```diff
@@ -35,3 +35,5 @@
     (0.6243159222654036, -0.14191423819668741),
     (0.563207308766221, 0.5152506858187523),
 ]
+
+MEASUREMENT_MODES = ["mass", "volume"]
```
yangchoo, line 39: nit: Let's rename this to `PROVISION_MEASUREMENT_MODES`, since there are other measurement modes (e.g. Spectrophotometry)

autoprotocol/protocol.py
```diff
@@ -5047,9 +5047,9 @@ def provision(self, resource_id, dests, volumes):
         dests : Well or WellGroup or list(Well)
           Destination(s) for specified resource.
         volumes : str or Unit or list(str) or list(Unit)
-          Volume(s) to transfer of the resource to each destination well.  If
-          one volume of specified, each destination well recieve that volume of
-          the resource.  If destinations should recieve different volumes, each
+          Volume(s) or Mass(es) to transfer of the resource to each destination well.  If
+          one volume/mass is specified, each destination well receive that volume/mass of
```
yangchoo, line 5051: nit: think you missed the `/` here

autoprotocol/protocol.py
```diff
@@ -5047,9 +5047,9 @@ def provision(self, resource_id, dests, volumes):
         dests : Well or WellGroup or list(Well)
           Destination(s) for specified resource.
         volumes : str or Unit or list(str) or list(Unit)
```
yangchoo, line 5049: I understand there was the comment with @joshhacksthings on this... to use `volumes` rather than `amounts` for backwards compat. 

However, having `volumes` as the docstring/argument name seems pretty unintuitive. One possible workaround to support backwards compat would be to add volumes as a kwarg (keyword arugment).
I.e. `provision(self, resource_id, dests, amounts, volumes=None)`

Hmm.. for now we can add a comment that it's for backwards compat/will be deprecated. Forward thinking, we could look into something similar to https://pypi.org/project/Deprecated/ as well. That would be nice for marking these.

What do you think, @joshhacksthings ?

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,30 @@ def provision(self, resource_id, dests, volumes):
...
+                    )
                 )
         return provision_instructions_to_return
 
+    def _identify_measurement_mode(self, volumes):
+        unique_measure_modes = set()
+        for amount in volumes:
+            if not isinstance(amount, (str, Unit)):
+                raise TypeError(f"Provided amount {amount} is not supported.")
+            if amount.unit.endswith("liter"):
```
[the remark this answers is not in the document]
ramu-phanimukla: @yangchoo & @joshhacksthings 
I prefer using below as it will be more generic:
```
# For mass
amount.dimensionality == Unit(1, "gram").dimensionality 
# For volume
amount.dimensionality == Unit(1, "liter").dimensionality
```
Modified the code accordingly

autoprotocol/protocol.py
ramu-phanimukla: Done

autoprotocol/constants.py
ramu-phanimukla: Renamed.

autoprotocol/constants.py
```diff
@@ -35,3 +35,5 @@
     (0.6243159222654036, -0.14191423819668741),
     (0.563207308766221, 0.5152506858187523),
 ]
+
+PROVISION_MEASUREMENT_MODES = ["mass", "volume"]
```
joshhacksthings, line 39: nit: could we move this into the `Provision` Instruction? or the protocol method?

autoprotocol/instruction.py
```diff
@@ -9,6 +9,7 @@
 
 # pragma pylint: disable=too-few-public-methods, redefined-builtin
 from .builders import *  # pylint: disable=unused-wildcard-import
+from .constants import *  # pylint: disable=unused-wildcard-import
```
joshhacksthings, line 12: nit: if we know for sure we're only importing one object (`PROVISION_MEASUREMENT_MODES`) I think it would be worth explicitly importing it, and avoiding a wildcard import

autoprotocol/instruction.py
```diff
@@ -1385,9 +1388,19 @@ class Provision(Instruction):
 
     """
 
-    def __init__(self, resource_id, dests):
+    def __init__(self, resource_id, dests, measurement_mode="volume"):
+        if measurement_mode not in PROVISION_MEASUREMENT_MODES:
+            raise RuntimeError(
```
joshhacksthings, line 1393: add `RuntimeError` to the "Raises" section of this function's docstring

docs/changelog.rst
```diff
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :support:`255` Add support for provisioning mass based resources
```
joshhacksthings, line 5: nit: I'd argue this could be a `feature`

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,31 @@ def provision(self, resource_id, dests, volumes):
...
+                    self._append_and_return(
+                        Provision(resource_id, dest_group, measurement_mode)
+                    )
                 )
         return provision_instructions_to_return
 
+    def _identify_measurement_mode(self, volumes):
+        unique_measure_modes = set()
+        for amount in volumes:
+            print(amount.dimensionality)
```
joshhacksthings, line 5153: remove `print`

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,31 @@ def provision(self, resource_id, dests, volumes):
...
             else:
                 provision_instructions_to_return.append(
-                    self._append_and_return(Provision(resource_id, dest_group))
+                    self._append_and_return(
+                        Provision(resource_id, dest_group, measurement_mode)
+                    )
                 )
         return provision_instructions_to_return
 
+    def _identify_measurement_mode(self, volumes):
```
joshhacksthings, line 5150: nit: rename `volumes` argument to something more descriptive. It looks like we're expecting a list of `Unit`s. I'd be fine with `amounts` or something like that

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,31 @@ def provision(self, resource_id, dests, volumes):
...
+        unique_measure_modes = set()
+        for amount in volumes:
+            print(amount.dimensionality)
+            if not isinstance(amount, (str, Unit)):
```
joshhacksthings, line 5154: this check is a bit unwieldy. 

First, we should not be allowing `amount` to be of type `str` as the subsequent steps call an attribute that only `Unit` objects have.

Second, we could use something like `amount = parse_unit(amount)` to sanely parse the unit provided. It will guarantee `amount` is type `Unit`

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,31 @@ def provision(self, resource_id, dests, volumes):
...
+            if amount.dimensionality == Unit(1, "liter").dimensionality:
+                unique_measure_modes.add("volume")
+            elif amount.dimensionality == Unit(1, "gram").dimensionality:
+                unique_measure_modes.add("mass")
+            else:
+                raise ValueError(
+                    f"Provisioning of resources with measurement unit of {amount.unit} is not supported."
+                )
+        if len(unique_measure_modes) != 1:
+            raise ValueError("Received amounts with more than one measurement mode")
```
joshhacksthings, line 5165: Up until this line, this method has been instruction agnostic. ie, you could use `_identify_measurement_mode` to decipher a measurement mode from any list of `Unit`s provided.

Starting with this `else` block, we raise a `ValueError` based on the `provision` method's support of a type of unit. I think we should make this method clearer, maybe renaming to `_identify_provision_mode`, so it's clear this is a provision-specific function

autoprotocol/protocol.py
```diff
@@ -5135,10 +5141,31 @@ def provision(self, resource_id, dests, volumes):
...
+        measurement_mode = unique_measure_modes.pop()
```
joshhacksthings, line 5166: nit: would this be a good place to check that the single item in `unique_measure_mode` is also in `PROVISION_MEASUREMENT_MODES`?

test/data/provision_for_mass.json
joshhacksthings: Got it, makes total sense. Since we can have these keys in any order, can we move `measurement_mode` to follow the line `"op": "provision",` just to harmonize with other provision json files?

test/protocol_test.py
```diff
@@ -2738,9 +2738,9 @@ def test_convert_provision(self):
         p1 = Protocol()
         c1 = p1.ref("c1", id=None, cont_type="96-pcr", discard=True)
         p1.incubate(c1, where="ambient", duration="1:hour", uncovered=True)
-        p1.provision("rs18s8x4qbsvjz", c1.well(0), volumes="10:microliter")
+        p1.provision("rs18s8x4qbsvjz", c1.well(0), "10:microliter")
         p1.incubate(c1, where="ambient", duration="1:hour", uncovered=True)
-        p1.provision("rs18s8x4qbsvjz", c1.well(0), volumes="10:microliter")
+        p1.provision("rs18s8x4qbsvjz", c1.well(0), "10:microliter")
```
joshhacksthings, line 2743: let's keep the `volumes` kwarg, just to ensure backwards compat

test/provision_test.py
```diff
@@ -174,7 +174,7 @@ def test_for_multiple_dispenses_of_resource_in_containers_larger_than_900ml(self
             .well(0)
             .set_volume("2:microliter")
         )
-        p.provision("rs17gmh5wafm5p", w1, "1500:microliter")
+        p.provision("rs17gmh5wafm5p", w1, volumes="1500:microliter")
```
joshhacksthings, line 177: same as above, let's keep the volumes kwarg, just to ensure backwards compat

test/provision_test.py
```diff
@@ -183,3 +183,45 @@ def test_for_multiple_dispenses_of_resource_in_containers_larger_than_900ml(self
             "split_provisions_by_volume.json"
         )
         assert expected_instruction_as_json == actual_instruction_as_json
+
+    def test_provision_well_with_mass(self):
+        p = Protocol()
```
joshhacksthings, line 188: I'd strongly recommend writing these `Protocol` and `Well` objects as pytest fixtures, just to cutdown on rewriting boilerplate

a good example can be found in [TestTransferVolume](https://github.com/autoprotocol/autoprotocol-python/blob/3d09f1271e3695a260b0ce4e85a8445a3ef38b8e/test/protocol_test.py#L3258)

autoprotocol/constants.py
ramu-phanimukla: As they are really serving as constants and we have separate file for that, I feel it is the right place.

test/provision_test.py
ramu-phanimukla: Updated `provision_test.py` tests to use `pytest fixtures`

test/provision_test.py
ramu-phanimukla: Done

test/protocol_test.py
ramu-phanimukla: Done

autoprotocol/protocol.py
ramu-phanimukla: I did thought about it, but it will be redundant as we already have that check in the `Provision` constructor.

docs/changelog.rst
ramu-phanimukla: Makes sense. Modified to:
```
* :feature:`255` Add support for provisioning of resources by mass
```

autoprotocol/instruction.py
ramu-phanimukla: Updated `Raises` section

autoprotocol/instruction.py
ramu-phanimukla: Followed previous import statement. Anyhow fixed it.

autoprotocol/instruction.py
```diff
@@ -1385,9 +1388,17 @@ class Provision(Instruction):
 
     """
 
-    def __init__(self, resource_id, dests):
+    def __init__(self, resource_id, dests, measurement_mode="volume"):
+        if measurement_mode not in MEASUREMENT_MODES:
+            raise RuntimeError(f"{measurement_mode} is not a valid measurement mode")
```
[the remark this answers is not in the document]
ramu-phanimukla: Added `If the measurement mode is not supported.` to "Raises" section of the docstring.

ramu-phanimukla:
@joshhacksthings & @yangchoo All comments addressed

test/data/provision_for_mass.json
ramu-phanimukla: Ensured same order of keys across all files

autoprotocol/protocol.py
```diff
@@ -5034,7 +5034,7 @@ def image_plate(self, ref, mode, dataref):
         """
         return self._append_and_return(ImagePlate(ref, mode, dataref))
 
-    def provision(self, resource_id, dests, volumes):
+    def provision(self, resource_id, dests, volumes=None, amounts=None):
```
yangchoo, line 5037: Would prefer `def provision(self, resource_id, dests, amounts, volumes=None)` instead. This retains backwards compat support while allowing the user to provide mass as the third argument. I.e. `provision(..., ..., Unit("1:gram"))`

yangchoo, line 5037: would prefer to swap `volumes` and `amounts` positions, so it'll allow someone to specify mass as the third positional arg

autoprotocol/protocol.py
ramu-phanimukla: Done

autoprotocol/protocol.py
ramu-phanimukla: Renamed to `_identify_provision_mode`

autoprotocol/protocol.py
ramu-phanimukla: I see below code is redundant:
```
            if not isinstance(amount, (str, Unit)):
                raise TypeError(f"Provided amount {amount} is not supported.")
```
because it is handled by below code which is before `_identify_measurement_mode ` call:
```
        if not isinstance(provision_amounts, list):
            provision_amounts = [Unit(provision_amounts)] * len(dests)
        else:
            if len(provision_amounts) != len(dests):
                raise RuntimeError(
                    "To provision a resource into multiple "
                    "destinations with multiple volumes or masses, the  "
                    "list of volumes or masses must correspond with the "
                    "destinations in length and in order."
                )
            provision_amounts = [Unit(v) for v in provision_amounts]
```

autoprotocol/protocol.py
ramu-phanimukla: Renamed `provision_amounts`

autoprotocol/protocol.py
yangchoo: done, thanks!

autoprotocol/protocol.py
```diff
@@ -5046,85 +5046,109 @@ def provision(self, resource_id, dests, volumes):
           Resource ID from catalog.
         dests : Well or WellGroup or list(Well)
           Destination(s) for specified resource.
+        amounts : str or Unit or list(str) or list(Unit)
+          Mass(es) to transfer of the resource to each destination well.  If
+          one mass is specified, each destination well receive that mass of
+          the resource.  If destinations should receive different mass, each
```
yangchoo, line 5052: Should this be `Volumes or Masses`?

autoprotocol/protocol.py
yangchoo: Ah I see. Its already been cast by the time it reaches here. 

In that case, I think its fine to remove the `str` type-check. I.e. just ` if not isinstance(amount, Unit)`

autoprotocol/protocol.py
joshhacksthings: Yeah, I was thinking that the check here would be in place of that one. I'll leave it up to you

autoprotocol/protocol.py
```diff
@@ -5046,85 +5046,109 @@ def provision(self, resource_id, dests, volumes):
...
-                    "list of volumes must correspond with the "
+                    "destinations with multiple volumes or masses, the  "
+                    "list of volumes or masses must correspond with the "
                     "destinations in length and in order."
                 )
-            volumes = [Unit(v) for v in volumes]
-        for v in volumes:
-            if not isinstance(v, (str, Unit)):
-                raise TypeError("Volume must be a string or Unit.")
+            provision_amounts = [Unit(v) for v in provision_amounts]
```
yangchoo, line 5116: missed this from the previous review, but fwiw, you can do `[parse_unit(v, ["liter", "gram") for v in provision_amounts]` instead of `[Unit(v) for v in provision_amounts]`

That way, there'll be additional type checks earlier on. Its not expensive and wraps it with nicer error messages

autoprotocol/instruction.py
```diff
@@ -9,6 +9,7 @@
 
 # pragma pylint: disable=too-few-public-methods, redefined-builtin
 from .builders import *  # pylint: disable=unused-wildcard-import
+from .constants import PROVISION_MEASUREMENT_MODES
```
joshhacksthings, line 12: pasting conversation from DMs here:

Ramu: 
I was following the previous import statement and if I make change, we might have inconsistency. What do you think?
me:
The import above this (from `.builders` is fine using a wildcard import because the code in `protocol.instruction` depends on all 11 classes defined in `.builders`.
We only need a single object, `PROVISION_MEASUREMENT_MODES` from `.constants`.

autoprotocol/protocol.py
```diff
@@ -5046,85 +5046,109 @@ def provision(self, resource_id, dests, volumes):
...
         volumes : str or Unit or list(str) or list(Unit)
-          Volume(s) to transfer of the resource to each destination well.  If
-          one volume of specified, each destination well recieve that volume of
-          the resource.  If destinations should recieve different volumes, each
+          Volume to transfer of the resource to each destination well.  If
+          one volume is specified, each destination well receive that volume of the resource.
+          If destinations should receive different volumes, each
           one should be specified explicitly in a list matching the order of the
           specified destinations.
+          Note: If value is provided for this field, value for 'amounts' must be skipped
```
joshhacksthings, line 5062: nit: I'd rephrase to something like `volumes` and `amounts` arguments are mutually exclusive. Only one is required.

autoprotocol/protocol.py
```diff
@@ -5046,85 +5046,109 @@ def provision(self, resource_id, dests, volumes):
...
         dests = WellGroup(dests)
+
         if not isinstance(resource_id, str):
             raise TypeError("Resource ID must be a string.")
-        if not isinstance(volumes, list):
-            volumes = [Unit(volumes)] * len(dests)
+
+        if (volumes is None and amounts is None) or (volumes and amounts):
+            raise ValueError(
+                "Either volumes or amounts should have value/s, but not both."
```
joshhacksthings, line 5098: nit: change "value/s" to "value(s)"

autoprotocol/protocol.py
joshhacksthings: Yep

test/protocol_test.py
```diff
@@ -2740,7 +2740,7 @@ def test_convert_provision(self):
         p1.incubate(c1, where="ambient", duration="1:hour", uncovered=True)
         p1.provision("rs18s8x4qbsvjz", c1.well(0), volumes="10:microliter")
         p1.incubate(c1, where="ambient", duration="1:hour", uncovered=True)
-        p1.provision("rs18s8x4qbsvjz", c1.well(0), volumes="10:microliter")
+        p1.provision("rs18s8x4qbsvjz", c1.well(0), "10:microliter")
```
joshhacksthings, line 2743: nit: I'd keep the `volumes=` key word here since the behavior should remain unchanged

test/provision_test.py
```diff
@@ -88,21 +80,15 @@ def test_with_consecutive_repeated_wells(self):
         assert expected_instruction_as_json == actual_instruction_as_json
 
     def test_with_repeated_wells_but_discontinious(self):
```
joshhacksthings, line 82: nit: correct typo `discontinious` to `discontinuous` (since we're changing this method I feel it's within scope)

test/provision_test.py
```diff
@@ -168,18 +154,46 @@ def test_provision_with_sealed_container(self):
...
-            p.as_dict()["instructions"], indent=2, sort_keys=True
+            self.p.as_dict()["instructions"], indent=2, sort_keys=True
         )
         expected_instruction_as_json = TestUtils.read_json_file(
             "split_provisions_by_volume.json"
         )
         assert expected_instruction_as_json == actual_instruction_as_json
+
+    def test_provision_well_with_mass(self):
+        self.p.provision("rs17gmh5wafm5p", self.w1, amounts="50:ug")
```
joshhacksthings, line 174: nit: we're missing a test case that correctly provisions more than one mass

autoprotocol/protocol.py
ramu-phanimukla: If someone is trying to create provision object like :
```
provision = Provision(resource_id, dest_group, measurement_mode)
```
the check in the constructor will prevent invalid measurement_mode being set in the object

autoprotocol/protocol.py
ramu-phanimukla: Corrected the documentation

autoprotocol/protocol.py
ramu-phanimukla: Fixed

autoprotocol/instruction.py
```diff
@@ -1380,14 +1383,25 @@ class Provision(Instruction):
     RuntimeError
       If length of the list of volumes specified does not match the number of
       destination wells specified.
+      If the measurement mode is not supported.
```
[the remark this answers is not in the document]
ramu-phanimukla: Fixed

autoprotocol/protocol.py
yangchoo: thanks!

autoprotocol/protocol.py
ramu-phanimukla: Fixed

autoprotocol/protocol.py
ramu-phanimukla: Fixed

test/protocol_test.py
ramu-phanimukla: Fixed

test/provision_test.py
ramu-phanimukla: Fixed

test/provision_test.py
ramu-phanimukla: Added explicit test case `test_provision_multiple_wells_with_diff_masses`

docs/changelog.rst
```diff
@@ -2,6 +2,8 @@
 Changelog
 =========
 
+* :feature:`255` Add support for provisioning of resources by mass
```
yangchoo, line 5: this should be 254

test/protocol_test.py
joshhacksthings: still missing `volumes`

original line is:
`p1.provision("rs18s8x4qbsvjz", c1.well(0), volumes="10:microliter")`

PR #255, opened 2020-06-02
fix: compatibility with txappy

aota001:
fix: desired_mode
minimal_transfer in txappy returns None for `mode_params`. Updating ApPy to make it compatible.

PR #256, opened 2020-06-02
v7.0.1

aota001:
Release v7.0.1 with a bug fix from #255

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,7 @@ Changelog
 =========
 
 * :release:`7.0.0 <2020-05-28>`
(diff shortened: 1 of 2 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :release:`7.0.1 <2020-06-02>`
+* :bug:`255 major` Update desired_mode method in LiquidHandle to allow for cases where mode_param is empty
```
yangchoo, line 6: if we're releasing right away with a bugfix, we can use `255` without the major flag

aota001: updated

PR #258, opened 2020-06-15
feat: assign time_constraints to instance attribute

aota001:
`time_constraints pretty` was using a has then set pattern. Instead, we'll assign `time_constraint` value (list) to the attribute to be inherited by Protocol instances.

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -889,8 +889,14 @@ class Protocol(object):
             for a in dir(self)
             if not a.startswith("__") and not callable(getattr(self, a))
         ]
-
-        explicit_props = ["outs", "refs", "instructions", "time_constraints"]
+        # if either refs or instructions is empty raise error
+        if not self.refs or not self.instructions:
+            raise RuntimeError("there must be at least one instruction and a ref")
+        # time_constraints is an optional attribute and should not be serialized if value is []
+        if not self.time_constraints:
+            explicit_props = ["outs", "refs", "instructions"]
+        else:
+            explicit_props = ["outs", "refs", "instructions", "time_constraints"]
 
         return {
             attr: self._refify(getattr(self, attr))
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -896,11 +896,14 @@ class Protocol(object):
         # if either refs or instructions is empty raise error
         if not self.refs or not self.instructions:
             raise RuntimeError("there must be at least one instruction and a ref")
-        # time_constraints is an optional attribute and should not be serialized if value is []
-        if not self.time_constraints:
-            explicit_props = ["outs", "refs", "instructions"]
-        else:
-            explicit_props = ["outs", "refs", "instructions", "time_constraints"]
+
+        explicit_props = ["outs", "refs", "instructions"]
+        # attributes that are always serialized.
+        optional_props = ["time_constraints"]
+        # optional attributes that are serialized only when there are values
+        for prop in optional_props:
+            if getattr(self, prop):
+                explicit_props.append(prop)
 
         return {
             attr: self._refify(getattr(self, attr))
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -499,13 +499,6 @@ class Protocol(object):
 
         state_strings = ["start", "end"]
 
-        def add_time_constraint_internal(time_const):
-            setattr(
-                self,
-                "time_constraints",
-                (getattr(self, "time_constraints", []) + [time_const]),
-            )
-
         keys = []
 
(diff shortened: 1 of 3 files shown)


yangchoo (changes requested):
We should change `https://github.com/autoprotocol/autoprotocol-python/blob/master/autoprotocol/protocol.py#L501` so that it uses the `time_constraints` attribute, and change all references to `add_time_constraint_internal`

aota001:
@yangchoo removed `add_time_constraint_internal`. (Is this what you meant?)

test/protocol_test.py
```diff
@@ -448,9 +448,6 @@ def test_time_constraint(self, dummy_protocol):
         p.cover(plate_2)
         time_point_2 = p.get_instruction_index()
 
-        with pytest.raises(AttributeError):
-            p.time_constraints  # pylint: disable=pointless-statement
-
```
yangchoo, line 453: fyi: I believe the main intent here was to ensure we don't unnecessarily serialize the field if its empty.

Something you could do is to replicate a check for that by examining the `p.as_dict` output, and ensuring the `time_constraints` key is not present.

yangchoo (dismissed):
yup, that's exactly what I meant. Thanks!

Please
- [x] Add to the changelog
- [x] Consider replicating `time_constraints` empty check in `p.as_dict()` output

Other than that, this looks good to me.

autoprotocol/protocol.py
```diff
@@ -126,11 +128,18 @@ class Protocol(object):
...
+        refs=None,
+        instructions=None,
+        propagate_properties=False,
+        time_constraints=None,
+    ):
         super(Protocol, self).__init__()
         self.refs = refs or {}
         self.instructions = instructions or []
         self.propagate_properties = propagate_properties
+        self.time_constraints = time_constraints or []
```
yangchoo, line 142: sorry going back on this, I think it should default to None, see below

yangchoo (changes requested):
Requesting changes to figure out the serialization issue. See comments

test/protocol_test.py
```diff
@@ -448,8 +448,7 @@ def test_time_constraint(self, dummy_protocol):
         p.cover(plate_2)
         time_point_2 = p.get_instruction_index()
 
-        with pytest.raises(AttributeError):
-            p.time_constraints  # pylint: disable=pointless-statement
+        assert not p.as_dict()["time_constraints"]
```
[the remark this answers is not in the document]
aota001: Thanks for the clarification! I was little confused on what behavior we want from the previous comments. :p Will update :)

autoprotocol/protocol.py
aota001: @yangchoo Actually, if we set this to None instead of [], either we'll need to add extra checks in LN#608, 615 and 624, or re-set the attribute from None to [] so we can have +=[...] operation. 
Kept it to [] for now, but can change it to None if we really want that...

autoprotocol/protocol.py
```diff
@@ -888,8 +889,14 @@ def as_dict(self):
...
-
-        explicit_props = ["outs", "refs", "instructions", "time_constraints"]
+        # if either refs or instructions is empty raise error
+        if not self.refs or not self.instructions:
+            raise RuntimeError("there must be at least one instruction and a ref")
+        # time_constraints is an optional attribute and should not be serialized if value is []
+        if not self.time_constraints:
+            explicit_props = ["outs", "refs", "instructions"]
+        else:
+            explicit_props = ["outs", "refs", "instructions", "time_constraints"]
```
aota001, line 899: @yangchoo Added check to exclude 'time_constraints' if empty. 
Also added RuntimeError if either refs or instructions is empty, but did we want this in the end? Or did you want to just serialize []? (Wasn't 100% sure, now that time_constraints is set to []...)

autoprotocol/protocol.py
yangchoo: right.. I see. Okay then. Seems acceptable, at the cost of special casing required downstream

autoprotocol/protocol.py
yangchoo: oof. Yea... this is ugly. 

I think this is fine. I would clean this up a little, by making this distinction clearer between things we always serialize even when empty or not

E.g.
```
explicit_props = ["outs", "refs", "instructions"]  # things we always serialize
optional_props = ["time_constraints"]  # things we serialize if it is not empty

for prop in optional_props:
  if not prop:
    explicit_prop.append(prop)
```

yangchoo (dismissed):
looks good to me. Please see cleanup comment.

ramu-phanimukla (approved):
Based on @yangchoo latest comment there are no further changes required/requested and hence approving.

PR #263, opened 2020-07-30
feat: parse compounds from manifest in aliquots

aota001:
* Add `compounds` to Well attributes in container.py
* Add feature to append `compounds` in harness.py
* Update all tests

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -8,6 +8,7 @@ Container, Well, WellGroup objects and associated functions
 """
 
 import json
+import re
 import warnings
 
 from .constants import SBS_FORMAT_SHAPES
@@ -38,6 +39,7 @@ class Well(object):
         self.index = index
         self.volume = None
         self.name = None
+        self.compounds = None
         self.properties = {}
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


aota001:
This is still wip, but wanted to start sharing so that I can make corrections as I go along since I'm making changes in containers... thanks!

autoprotocol/container.py
```diff
@@ -168,6 +227,25 @@ def set_name(self, name):
         self.name = name
         return self
 
+    def set_compounds(self, compounds):
+        """
+            Set a name for this well for it to be included in a protocol's
```
joshhacksthings, line 232: nit: "Set a compound" not "name"

autoprotocol/container.py
```diff
@@ -207,7 +285,7 @@ def __repr__(self):
         Return a string representation of a Well.
 
         """
-        return f"Well({str(self.container)}, {str(self.index)}, " f"{str(self.volume)})"
+        return f"Well({str(self.container)}, {str(self.index)}, " f"{str(self.volume)}, {str(self.compounds)})"
```
joshhacksthings, line 288: is this something we'll want to return if `.humanize()` is called on the well?

aota001:
Thanks @joshhacksthings. I just put up ASC proposal in https://github.com/autoprotocol/ascs/pull/28. Will add this to the description once it is approved.

aota001:
Will close this PR until the parent ASC is resolved. Also, due to a change in the product requirement, I no longer need this feature at this moment.

PR #264, opened 2020-08-20
Container type service

samjoch:


autoprotocol/container_type.py
```diff
@@ -91,53 +66,28 @@ class ContainerType(
...
-            shortname,
-            col_count,
-            dead_volume_ul,
-            safe_min_volume_ul,
-            true_max_vol_ul,
-            vendor,
-            cat_no,
-            prioritize_seal_or_cover,
-        )
+    def __init__(self, shortname):
```
samjoch, line 69: Some ideas around managing `baseUrl`:

**Adding a new argument when initialize**
ContainerType('96-pcr')
ContainerType('96-pcr', 'https://secure.strateos.com/api/container_types')

**Environment variable**
baseUrl = ENV['CONTAINER_TYPE_URL'] || 'https://secure.strateos.com/api/container_types'

**Class argument**
ContainerType.baseUrl =  'https://secure.strateos.com/api/container_types'
ContainerType('96-pcr')

setup.py
```diff
@@ -14,6 +14,7 @@ class PyTest(TestCommand):
     def initialize_options(self):
         TestCommand.initialize_options(self)
         self.pytest_args = "--cov=autoprotocol --cov-report=term"  # pylint: disable=attribute-defined-outside-init
+        # -x test/container_type_test.py
```
yangchoo, line 17: should this be here?

test/conftest.py
```diff
@@ -1,263 +1,826 @@
...
 
 
+@pytest.fixture(scope="module", autouse=True)
+def run_around_tests():
+    print("Define mock")
+
+    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
+        rsps.add(
+            rsps.GET,
+            "https://secure.strateos.com/api/container_types/res-sw96-hp",
```
yangchoo, line 16: nit: this is a pretty huge function,  would prefer defining it in a separate file and importing it, or directly saving the response as a json file and deserialize it here

setup.py
samjoch: nope

samjoch:
@yangchoo 

> 1. Autoprotocol-Python now requires an internet connection (and availability of some ContainerType service) in order to function.

That is true. I will addressed the caching issue.

> 2. ContainerType attributes are entirely dynamic I.e. there are no guarantees on which attributes will be available other than shortname.

That is true. Are you saying we need to maintain a list of attributes here? Because what is the point of using a container type service and maintaining a static list of container type attributes? The service could provide it?

__Another thought__: Can we say that we don't even need to call a container type service in `autoprotocol-python`. A container type should be only used as a reference (it's `shortname`). Don't we expect from the scientist to know which container type they are using? We are missing an `autoprotocol` service where a protocol should be validated against. Thoughts?

yangchoo:
>That is true. Are you saying we need to maintain a list of attributes here? Because what is the point of using a container type service and maintaining a static list of container type attributes? The service could provide it?

My bias is towards having a versioned API where there are guarantees on the schema of the data. For example, something following an OpenAPI specification. That way, while the values may change, there are guarantees around the presence of certain  fields.

That said, I see where you're coming from in terms of making the ContainerType itself be more flexible. I.e. be able to ingest other ContainerType representations containing additional fields, potentially from other services.  

I personally prefer consistency over flexibility, but if we were to want to support that behavior, I was thinking we should at least mark certain fields as core/required. And those fields will be determined by the methods which are commonly called/defined in `container_type.py`.

So for example, `well_count`, `col_count` will be required, since those definitions are pretty central to the definition of a ContainerType.

PR #265, opened 2020-08-26
feat: Add harness support for units specified in ASC-51

yangchoo:
ASC-51 added formal support for concentration units (amount, mass,
volume) into the Autoprotocol spec.

This adds support for these units into the harness, keeping to a similar
namespace.

Note that support for `concentration(mass)`, `concentration(molar)`
should be deprecated in some future major version bump.


joshhacksthings (dismissed):
LGTM 🚀

joshhacksthings (approved):
looks good!

PR #267, opened 2020-09-03
style: Pin black to 20.8b1

yangchoo:
Previously we were running black on `stable`, but this occasionally led
to local/CI discrepancies when the cache is utilized. Pinning it can
help reduce these issues.


rhysormond (approved):
Lgtm!

PR #269, opened 2020-09-17
build: Fix Travis deployment issue

yangchoo:
We're currently spinning off multiple build jobs for all tests running
python 3.6.

We should only be spinning off a single deployment job, adding an
additional envvar filter for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ jobs:
   - python: '3.6'
     env: TOXENV=lint
   - python: '3.6'
-    env: TOXENV=clean,py36,stats_xml
+    env: TOXENV=clean,py36,stats_xml DEPLOY=true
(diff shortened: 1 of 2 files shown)


yangchoo:
From the last job, we see the following two issues
1. Pypi upload failing on the second/third try because we ran deployment three times. (e.g. https://travis-ci.org/github/autoprotocol/autoprotocol-python/builds/727889742). 
Solution: We should only run the deploy job on a single branch. Add an envvar to fix this.
2. Releases provider not working. `Skipping a deployment with the releases provider because this branch is not permitted: v7.2.0`. 
Solution: Adding a similar `tags: true` field to releases.

rhysormond (approved):
seems reasonable! thanks!

PR #270, opened 2020-09-18
acoustic transfer: raise an error if no transfers are generated

ericschade:
avoid situations where 0 uL transfers result in an empty "groups" field in autoprotocol and therefore cause an error when submitted to web. This change will cause the error to occur upstream of a web submission, notifying the protocol writer to avoid generating acoustic transfer instructions with 0 volume

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -1178,17 +1178,17 @@ class Protocol(object):
             if v > Unit(0, "microliter"):
                 transfers.append(xfer)
 
-        for x in transfers:
-            x["volume"] = round(x["volume"].to("nl"), max_decimal_places)
-
-        if transfers:
-            return self._append_and_return(
-                AcousticTransfer([{"transfer": transfers}], droplet_size)
-            )
-        else:
+        if not transfers:
             raise RuntimeError(
                 "At least one transfer must have a nonzero transfer volume."
             )
+
+        for x in transfers:
+            x["volume"] = round(x["volume"].to("nl"), max_decimal_places)
+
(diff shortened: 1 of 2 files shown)


autoprotocol/protocol.py
```diff
@@ -1181,9 +1181,14 @@ def acoustic_transfer(
...
-            AcousticTransfer([{"transfer": transfers}], droplet_size)
-        )
+        if transfers:
+            return self._append_and_return(
+                AcousticTransfer([{"transfer": transfers}], droplet_size)
+            )
+        else:
+            raise RuntimeError(
+                "At least one transfer must have a nonzero transfer volume."
+            )
```
joshhacksthings, line 1191: I think we should move this block to line 1180. and instead rewrite the conditional as `if not transfer`, and then raise the error. 

Then, we don't have to wrap the return in the `if` statement because that looks ugly

joshhacksthings (changes requested):
Looks good. I think I want the RuntimeError raised a bit earlier, so the last thing in the function will be the `return` statement. This will also minimize wrapping it a conditional.

TODOs:
- [x] move L1189-L1191 block to L1180
- [ ] wrap the L1180 block in `if not transfers`
- [ ] then remove the `if transfers` check in line 1184

autoprotocol/protocol.py
ericschade: the changes are made

joshhacksthings (dismissed):
Great! Forgot to mention previously this will need a changelog update, hen you're good to go!

PR #274, opened 2020-10-14
Stdv 333 container with wells bug

EribertoLopez:
STDV-333: https://strateos.atlassian.net/browse/STDV-333

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     rev: 20.8b1
     hooks:
     -   id: black
-        language_version: python3.6
+        language_version: python3
 -   repo: https://github.com/pycqa/pylint
     rev: pylint-2.5.2
     hooks:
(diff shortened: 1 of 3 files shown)


EribertoLopez:
@yangchoo I was having issues with the pre-commit environment that runs. I found that changing the black hook python version from `python3.6` to `python3` allowed the pre-commit environment to build and commit my change. Not sure if I'm missing something in the dependencies? Any suggestions would be helpful.

ericschade:
> @yangchoo I was having issues with the pre-commit environment that runs. I found that changing the black hook python version from `python3.6` to `python3` allowed the pre-commit environment to build and commit my change. Not sure if I'm missing something in the dependencies? Any suggestions would be helpful.

i just committed and didnt have any issues fwiw

ericschade (dismissed):
`is` -> `==` change looks good, not sure about the pre-commit config issue though

ericschade (approved):

PR #275, opened 2020-10-15
STDV 304 acoustic transfer one_source bug fix

ericschade:
well.volume -> well.available_volume to account for dead volume

branch should be STDV-304 not 340... oops

### DIFF
--- a/test/protocol_test.py
+++ b/test/protocol_test.py
@@ -1274,18 +1274,30 @@ class TestAcousticTransfer(object):
         p = dummy_protocol
         echo = p.ref("echo", None, "384-echo", discard=True)
         dest = p.ref("dest", None, "384-flat", discard=True)
+
+        # dead volume for source wells(15 uL) results in 0 available volume, not enough to transfer
+        with pytest.raises(RuntimeError):
+            p.acoustic_transfer(
+                echo.wells(0, 1).set_volume("2:microliter"),
+                dest.wells(0, 1, 2, 3),
+                "1:microliter",
+                one_source=True,
+            )
+
         p.acoustic_transfer(
-            echo.wells(0, 1).set_volume("2:microliter"),
-            dest.wells(0, 1, 2, 3),
-            "1:microliter",
+            echo.wells(0, 1, 2, 3).set_volume("60:microliter"),
+            dest.wells(list(range(0, 60))),
+            "2:microliter",
             one_source=True,
         )
+
         assert p.instructions[-1].data["groups"][0]["transfer"][-1][
             "from"
-        ] == echo.well(1)
+        ] == echo.well(2)
         assert p.instructions[-1].data["groups"][0]["transfer"][0]["from"] == echo.well(
             0
         )
+        assert echo.well(0).volume == echo.container_type.dead_volume_ul
 
     def test_droplet_size(self, dummy_protocol):
         p = dummy_protocol
--- a/test/protocol_test.py
+++ b/test/protocol_test.py
@@ -1270,34 +1270,39 @@ class TestAcousticTransfer(object):
         )
         assert len(p.instructions) == 4
 
-    def test_one_source(self, dummy_protocol):
+    @pytest.mark.parametrize(
+        "source_vol", ["2:microliter", "50:microliter", "17:microliter"]
+    )
+    @pytest.mark.parametrize("source_wells", [[0, 1, 2, 3], list(range(0, 60))])
+    @pytest.mark.parametrize(
+        "transfer_vol", ["1:microliter", "1.5:microliter", "2:microliter"]
+    )
+    def test_one_source(self, dummy_protocol, source_vol, source_wells, transfer_vol):
         p = dummy_protocol
         echo = p.ref("echo", None, "384-echo", discard=True)
         dest = p.ref("dest", None, "384-flat", discard=True)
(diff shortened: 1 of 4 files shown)


test/protocol_test.py
```diff
@@ -1274,18 +1274,30 @@ def test_one_source(self, dummy_protocol):
         p = dummy_protocol
         echo = p.ref("echo", None, "384-echo", discard=True)
         dest = p.ref("dest", None, "384-flat", discard=True)
+
```
EribertoLopez, line 1277: You're on the right track, but I think what would make this test stronger is if you used similar params in your test otherwise you may not be catching the edge cases and only finding the cases that make you test pass. Maybe try something like this out

```
    @pytest.mark.parametrize("source_volume", ['2:microliter', '60:microliter'])
    def test_one_source(self, dummy_protocol, source_volume):
        p = dummy_protocol
        echo = p.ref("echo", None, "384-echo", discard=True)
        dest = p.ref("dest", None, "384-flat", discard=True)

        one_source = echo.wells(0, 1, 2, 3).set_volume(source_volume)
        dests = dest.wells(0, 60)
        acoustic_transfer_vol = Unit('2:microliter')

        if sum([w.available_volume() for w in one_source]) < (len(dests)*acoustic_transfer_vol):
            with pytest.raises(RuntimeError):
                p.acoustic_transfer(
                    one_source,
                    dests,
                    acoustic_transfer_vol,
                    one_source=True
                )
        else:
            p.acoustic_transfer(
                one_source,
                dests,
                acoustic_transfer_vol,
                one_source=True
            )
            for w in one_source:
                assert w.volume >= w.container.container_type.dead_volume_ul
```

test/protocol_test.py
```diff
@@ -1274,18 +1274,30 @@ def test_one_source(self, dummy_protocol):
...
             one_source=True,
         )
+
         assert p.instructions[-1].data["groups"][0]["transfer"][-1][
             "from"
-        ] == echo.well(1)
+        ] == echo.well(2)
         assert p.instructions[-1].data["groups"][0]["transfer"][0]["from"] == echo.well(
             0
         )
```
EribertoLopez, line 1299: Do we need to know where the transfers came from or just that the volume was transferred and there was enough in the well? I think we may be able to get rid of these assertions.

EribertoLopez (changes requested):
On the right track, checkout my comments and feel free to reach out if you want to discuss things further.

test/protocol_test.py
ericschade: I definitely think this is a better way to structure the test so ill be pushing with that change, but im not sure what you mean by having "similar" params. which params should be similar to what?

test/protocol_test.py
ericschade: i think originally it was there to verify that the source wells were being changed if there was supposed to be a transfer from an empty well. in this case i guess its still doing something similar in that it is verifying that the second well will be used even though there is still volume remaining in the first source well

ericschade: but i think this could probably be tested more gracefully. the assertions are a bit clunky

test/protocol_test.py
EribertoLopez: The parameters to the instruction should be similar the only difference should be the volume in the source wells. In your OG test you were using different combination of wells for each case (ie: RuntimeError and no error). The test should be focusing on whether the change in source volume you made will raise an error when there isn't enough source volume.

ericschade: ok yeah, isolate one thing at a time to test. in the new commits i just threw in a couple extra things to test via parameterize for good measure

ericschade:
@EribertoLopez there was an intermediate commit which failed some of the precommit stuff including the black check, which i guess automatically reformatted things and didnt save the commit name so all of it is in the "unused variable fixed" commit.. also i thought maybe the black failure was related to the python version like you mentioned yesterday so i tried changing it but realized the issue was just the formatting so i changed it back. kinda messy commit history, i can do some cleaning up if it matters

EribertoLopez:
> @EribertoLopez there was an intermediate commit which failed some of the precommit stuff including the black check, which i guess automatically reformatted things and didnt save the commit name so all of it is in the "unused variable fixed" commit.. also i thought maybe the black failure was related to the python version like you mentioned yesterday so i tried changing it but realized the issue was just the formatting so i changed it back. the result is a kinda poorly named commit history, i can do some cleaning up if it matters

I think it's fine, but just make sure to update the `docs/changelog.rst` with the type of change and the number of the pull request (ie: #275 in this case)

ericschade:
> > @EribertoLopez there was an intermediate commit which failed some of the precommit stuff including the black check, which i guess automatically reformatted things and didnt save the commit name so all of it is in the "unused variable fixed" commit.. also i thought maybe the black failure was related to the python version like you mentioned yesterday so i tried changing it but realized the issue was just the formatting so i changed it back. the result is a kinda poorly named commit history, i can do some cleaning up if it matters
> 
> I think it's fine, but just make sure to update the `docs/changelog.rst` with the type of change and the number of the pull request (ie: #275 in this case)

got it, cool

EribertoLopez (dismissed):

ericschade:
the changelog required a conflict resolution, so i need another approval @EribertoLopez, sorry

EribertoLopez (approved):
LGTM! 🚀

PR #281, opened 2020-12-14
build: Add Downloads Badge

yangchoo:
Add downloads badge so that we can keep track of usage rate of this
package.


ramu-phanimukla (dismissed):
Format of the URL seems to be correct. 

@yangchoo Build is failing

yangchoo:
@ramu-phanimukla thanks! yea, seems like there's a missing dep for the docs. Added it, things should pass now.

ramu-phanimukla (approved):

PR #284, opened 2020-12-17
style: Automatically sort imports

yangchoo:
Adds an isort dependency for automatic import sorting. This is purely a
non-functional quality-of-life change.

### DIFF
--- a/test/provision_test.py
+++ b/test/provision_test.py
@@ -1,5 +1,7 @@
 import json
+
 from test.test_util import TestUtils
+
 import pytest
 
 from autoprotocol.protocol import Protocol
--- a/test/unit_test.py
+++ b/test/unit_test.py
@@ -1,6 +1,8 @@
-import pytest
-from autoprotocol.unit import Unit, UnitValueError
 from decimal import Decimal
+
+import pytest
+
+from autoprotocol.unit import Unit, UnitValueError
 
 
 class TestUnitType(object):
--- a/test/well_test.py
+++ b/test/well_test.py
@@ -1,6 +1,8 @@
 # pragma pylint: disable=missing-docstring,attribute-defined-outside-init
 import warnings
+
 import pytest
+
 from autoprotocol.container import Container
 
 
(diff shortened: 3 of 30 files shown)


test/provision_test.py
```diff
@@ -1,5 +1,7 @@
 import json
+
 from test.test_util import TestUtils
+
 import pytest
 
 from autoprotocol.protocol import Protocol
```
aota001, line 7: nit: sort these imports alphabetically as well?

test/unit_test.py
```diff
@@ -1,6 +1,8 @@
+from decimal import Decimal
+
 import pytest
```
aota001, line 3: shouldn't `import pytest` go on the first line?

test/well_test.py
```diff
@@ -1,6 +1,8 @@
 # pragma pylint: disable=missing-docstring,attribute-defined-outside-init
 import warnings
+
 import pytest
```
aota001, line 4: nit: sort imports alphabetically?

aota001 (approved):
few nits? :) otherwise lgtm

yangchoo:
@aota001 thanks for reviewing!

Generally, the order isort uses is
1. System imports
2. Library imports
3. Direct (from) imports and alphabetical.

Refer to https://github.com/PyCQA/isort for more configuration information.

test/provision_test.py
yangchoo: Yup, see my comment. In this case, since they separated into 3 groups:
1. system, python-level default libs -> 'test'
2. external python package -> 'pytest'
3. internal references -> autoprotocol

test/unit_test.py
yangchoo: Yup, see my comment. In this case, since they separated into 2 groups:
1. system, python-level default libs -> 'decimal'
2. external python package -> 'pytest'

This allows for easier distinguishing of sys vs. external imports

test/well_test.py
yangchoo: See above. Warnings is a system import

aota001 (approved):
@yangchoo thanks for the clarification!

PR #285, opened 2020-12-17
support: Add long description

yangchoo:
Our current pypi page has no long_description,
see https://pypi.org/project/autoprotocol/

Let's use our current README to populate that field.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,7 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


yangchoo:
Tested locally with `twine check` as well as using the readme_render package used by pypi: `python -m readme_renderer README.rst -o /tmp/README.html`

joshhacksthings (approved):

PR #288, opened 2021-01-20
feat: SEA-710 Add compound type

sreetejreddy:


autoprotocol/harness.py
```diff
@@ -200,6 +200,17 @@ def convert_param(protocol, val, type_desc):
                 f"The value supplied to input '{label}' (type aliquot++) is "
                 f"improperly formatted."
             )
+    elif type == "compound":
+        return str(val)
```
yangchoo, line 204: We probably want some sort of first-class Compound class representation in the future. Ideally, we can actually add some validation on this (e.g. SMILES string).

yangchoo (changes requested):
Approach generally seems reasonable, however, let's at least do the following:
1. Add tests.
2. Add changes to docs/changelog.rst
3. PR to Autoprotocol Spec

We should actually add an ASC for `Compound` to the Autoprotocol Spec for the `Derived Types`field.  See https://github.com/autoprotocol/ascs/blob/master/Specification/Definitions/types.md

autoprotocol/harness.py
dzleidig: Found this: https://gist.github.com/lsauer/1312860/264ae813c2bd2c27a769d261c8c6b38da34e22fb

dzleidig: Eh, I'm not confident in that regex. I'm ok with not validating for now.

dzleidig: > We probably want some sort of first-class Compound class representation in the future

Agreed. Let's make a `Compound` class. 

The `val` should be a `dict` with a `notation` (possible values: `SMILES`, but maybe `InChI` in the future) and a `value`, similar to the strategy we adopted in here: https://github.com/strateos/web/blob/master/app/controllers/api/v1/inventory_searches_controller.rb#L85

dzleidig (changes requested):

autoprotocol/harness.py
dzleidig: > We should actually add an ASC for Compound to the Autoprotocol Spec for the Derived Typesfield. See https://github.com/autoprotocol/ascs/blob/master/Specification/Definitions/types.md 

I'll make a small story for this. We should probably take this up internally, @yangchoo.

dzleidig:
This implementation depends on the outcome of the ASC: https://github.com/autoprotocol/ascs/pull/29/files.

docs/changelog.rst
```diff
@@ -472,3 +472,4 @@ Changelog
 
 * :release:`1.0.0 <2014-01-22>`
 * :feature:`-` initializing ap-py
+* :feature:`-` Add "Compound" to derived types
```
yangchoo, line 475: this should go in reverse order and be at the top of the file.  The `-` should correspond to your PR number, i.e. `288`.

autoprotocol/compound.py
```diff
@@ -0,0 +1,12 @@
...
+    Represents a single Compound
+
+    Parameters
+    ----------
+    inchi : Standard International Chemical Identifier
+
+    """
+
+    def __init__(self, inchi):
+        self.inchi = inchi
```
yangchoo, line 12: Could we add a simple validation here? There are some simple regex validations available online.

yangchoo (changes requested):
Almost there. Please fix the following
- [x] standardize on using the proper capitalization of inchi, i.e. `InChI`.
- [x] fix changelog entry
- [x] add basic inchi validation, regex is sufficient for now
- [x] add tests for inchi validation

yangchoo (dismissed):
lgtm thanks!

autoprotocol/compound.py
```diff
@@ -0,0 +1,29 @@
...
+    ----------
+    InChI : Standard International Chemical Identifier
+
+    """
+
+    def __init__(self, InChI):
+        if self.is_valid(InChI):
+            self.InChI = InChI
+        else:
+            raise CompoundError(f" {InChI} is not a valid InChI key")
```
yangchoo, line 18: nit: extra space after the `f"`

yangchoo (approved):
lgtm, thanks!

PR #290, opened 2021-02-11
feat: Add informatics attribute to Instruction

aota001:
Add new optional attribute `informatics`
* `informatics` defaults to None for backward compatibility
* `informatics` is serialized only when it is not empty
* update `provision` to include `informatics` as an optional param

### DIFF
--- a/autoprotocol/informatics.py
+++ b/autoprotocol/informatics.py
@@ -27,7 +27,7 @@ class Informatics:
         pass
 
     @abstractmethod
-    def validate(self, data):
+    def validate(self):
         pass
 
 
--- a/autoprotocol/informatics.py
+++ b/autoprotocol/informatics.py
@@ -10,7 +10,6 @@ Contains all the Autoprotocol Informatics objects
 from abc import abstractmethod
 
 from .compound import Compound
-from .container import Container, WellGroup
 from .util import is_valid_well
 
 
@@ -27,19 +26,20 @@ class Informatics:
     def as_dict(self):
         pass
 
+    @abstractmethod
+    def validate(self, all_wells):
+        pass
+
 
 class AttachCompounds(Informatics):
     """
-    informatics type attach_compounds is constructed as a list of dict detailing
-    new compounds
+    Informatics type attach_compounds is constructed as a list of dict detailing
+    new compounds associated with target wells
 
     Parameters
     ----------
-    data:
-        dict of informatics data
-    all_wells: WellGroup
-        All wells used in the instruction that are applicable for attach_compounds
-        to take effect on.
+    data: dict
+        dict of Informatics data
 
     Returns
     -------
@@ -51,45 +51,16 @@ class AttachCompounds(Informatics):
         data must be a dict
     TypeError
         wells must be Well, list of Well, or WellGroup
-    ValueError
-        wells must be part of wells or containers instruction operates on
     TypeError
         compounds must be a list
     TypeError
         each element in compounds must be Compound type
     """
 
-    def __init__(self, data: dict, all_wells):
-        # turn data into specific attributes for this type of informatics
-        if not isinstance(data, dict):
-            raise TypeError(f"informatics data: {data} must be provided in a dict.")
+    def __init__(self, data: dict):
 
         self.wells = data["wells"]
         self.compounds = data["compounds"]
-
-        # validate wells
-        if not is_valid_well(self.wells):
-            raise TypeError(
-                f"wells: {self.wells} must be Well, list of Well or WellGroup."
-            )
-        wells = WellGroup(self.wells)
-        # wells must be one or more of the wells or container instruction is operating on.
-        if isinstance(all_wells, Container):
-            all_wells = all_wells.all_wells()
-        else:
-            all_wells = WellGroup(all_wells)
-        for well in wells.wells:
-            if well not in all_wells.wells:
-                raise ValueError(
-                    f"informatics well: {wells} must be one of the wells used in this instruction."
-                )
-
-        # validate compounds
-        if not isinstance(self.compounds, list):
-            raise TypeError(f"compounds: {self.compounds} must be provided in a list.")
-        for compd in self.compounds:
-            if not isinstance(compd, Compound):
-                raise TypeError(f"compound: {compd} must be Compound type.")
 
         super().__init__()
 
@@ -117,3 +88,37 @@ class AttachCompounds(Informatics):
             "type": "attach_compounds",
             "data": {"wells": self.wells, "compounds": self.compounds},
         }
+
+    def validate(self):
+        """
+        validate input dict has valid parameters to instantiate AttachCompounds
+
+        Parameters
+        ----------
+        data: dict
+            Informatics data
+
+        Returns
+        -------
+        data: dict
+            Validated dict of data
+
+        Raises
+        ------
+        TypeError
+            wells is a valid well type
+        TypeError
+            compounds is a list
+        TypeError
+            compounds element is a Compound
+        """
+        if not is_valid_well(self.wells):
+            raise TypeError(
+                f"wells: {self.wells} must be Well, list of Well or WellGroup."
+            )
+
+        if not isinstance(self.compounds, list):
+            raise TypeError(f"compounds: {self.compounds} must be provided in a list.")
+        for compd in self.compounds:
+            if not isinstance(compd, Compound):
+                raise TypeError(f"compound: {compd} must be Compound type.")
--- a/autoprotocol/informatics.py
+++ b/autoprotocol/informatics.py
@@ -108,9 +108,7 @@ class AttachCompounds(Informatics):
             )
 
         if not isinstance(self.compounds, list):
-            raise TypeError(
-                f"compounds: {self.compounds} must be provided in a list."
-            )
+            raise TypeError(f"compounds: {self.compounds} must be provided in a list.")
         for compd in self.compounds:
             if not isinstance(compd, Compound):
                 raise TypeError(f"compound: {compd} must be Compound type.")
--- a/autoprotocol/informatics.py
+++ b/autoprotocol/informatics.py
@@ -8,6 +8,7 @@ Contains all the Autoprotocol Informatics objects
 """
 
 from abc import abstractmethod
+
 from .compound import Compound
 from .container import Container, WellGroup
 from .util import is_valid_well
@@ -18,6 +19,7 @@ class Informatics:
     Base class for informatics attribute in an instruction that is to later be
     encoded as JSON.
     """
+
     def __init__(self):
         pass
 
@@ -56,12 +58,11 @@ class AttachCompounds(Informatics):
     TypeError
         each element in compounds must be Compound type
     """
+
     def __init__(self, data: dict, all_wells):
         # turn data into specific attributes for this type of informatics
         if not isinstance(data, dict):
-            raise TypeError(
-                f"informatics data: {data} must be provided in a dict."
-            )
+            raise TypeError(f"informatics data: {data} must be provided in a dict.")
 
         self.wells = data["wells"]
         self.compounds = data["compounds"]
@@ -85,34 +86,32 @@ class AttachCompounds(Informatics):
 
         # validate compounds
         if not isinstance(self.compounds, list):
-            raise TypeError(
-                f"compounds: {self.compounds} must be provided in a list."
-            )
+            raise TypeError(f"compounds: {self.compounds} must be provided in a list.")
         for compd in self.compounds:
             if not isinstance(compd, Compound):
                 raise TypeError(f"compound: {compd} must be Compound type.")
 
     def as_dict(self):
         """generates a Python object representation of Informatics attribute in
-           class Instruction
+        class Instruction
 
-            Returns
-            -------
-            dict
-                a dict of python objects that have the same structure as the
-                Autoprotocol JSON for the Informatics
+         Returns
+         -------
+         dict
+             a dict of python objects that have the same structure as the
+             Autoprotocol JSON for the Informatics
 
-            Notes
-            -----
-            Used as a part of JSON serialization of the Instruction
+         Notes
+         -----
+         Used as a part of JSON serialization of the Instruction
 
-            See Also
-            --------
-            :class:`Instruction` : Instruction class
+         See Also
+         --------
+         :class:`Instruction` : Instruction class
 
-         """
+        """
 
         return {
             "type": "attach_compounds",
-            "data": {"wells": self.wells, "compounds": self.compounds}
+            "data": {"wells": self.wells, "compounds": self.compounds},
         }
--- a/autoprotocol/informatics.py
+++ b/autoprotocol/informatics.py
@@ -1,0 +1,118 @@
+"""
+Contains all the Autoprotocol Informatics objects
+
+    :copyright: 2020 by The Autoprotocol Development Team, see AUTHORS
+        for more details.
+    :license: BSD, see LICENSE for more details
+
+"""
+
+from abc import abstractmethod
+from .compound import Compound
+from .container import Container, WellGroup
+from .util import is_valid_well
+
+
+class Informatics:
+    """
+    Base class for informatics attribute in an instruction that is to later be
+    encoded as JSON.
+    """
+    def __init__(self):
+        pass
+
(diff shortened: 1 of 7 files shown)


autoprotocol/protocol.py
```diff
@@ -654,6 +655,117 @@ def get_instruction_index(self):
...
+            Validated list of dict with informatics details
+
+        Raises
+        ------
+        TypeError
+            informatics must be provided in a list
+        TypeError
+            informatics element must be in dict
+        KeyError
+            informaticss dict must have correct keys
```
scottyantipa, line 684: typo `s/informaticss/informatics`

autoprotocol/protocol.py
```diff
@@ -654,6 +655,117 @@ def get_instruction_index(self):
...
+        all_wells: WellGroup
+            All wells used in the instruction that are applicable for informatics
+            to take effect on.
+
+        Returns
+        -------
+        List
+            Validated list of dict with informatics details
+
+        Raises
```
scottyantipa, line 677: Right now you're raising an error when you encounter the first issue. So the user (I think?) will only get information about a single error at a time. They'll fix it, re-submit, see the next error, etc. If thats the case I would recommend not throwing an error, and continuing to inspect all elements in the array (where possible) and return the full list of errors to the user at once.

aota001: Ah. Good point. Will update error handling :) thanks!

aota001: Hm. Looked into this, but couple of things...
1. most of the errors are built on top of each other, i.e. if the first error fails, the rest is uncheckable/does not make sense to run, so it makes sense to raise error at that point. 
2. the rest of the implementation seems to rely on raising exceptions to propagate error, which are eventually handled and output as a json with error message. It may be possible to append multiple errors (instead of raising exception), but it's not clear to me whether it can be done cleanly without impacting how error is handled at the top...? Also, I think the error propagation and handling mechanism should be consistent across the board. This is reasonable suggestion but I think it warrants additional discussion outside the scope of this change.

aota001:
Couple of things to note... There are some work around to achieve check to make sure informatics `wells` are part of `wells` or `refs` that instruction is operating on. 
1. Check is executed at the protocol level where the information on instruction target wells/refs are available. And because of that, I added a check to each instruction that expects `informatics` (this includes instructions in https://github.com/strateos/txappy/pull/42).
2. Initial idea was to add a validate abstract method in the base Informatics class, so that it can enforce all the subclasses to run `validate` upon instantiation, but I was not sure if `all_wells` would _always_ be required... so added check in the `AttachCompounds` for now. (Hence, Informatics base class does not have much contents to it at the moment.)
3. I think the check on wells against all wells in the instruction is necessary to avoid user errors at downstream. We definitely do not want to propagate effects on wrong wells/containers by accident. However, if there is a more elegant way to implement this, I'm open for suggestions...

autoprotocol/informatics.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class AttachCompounds(Informatics):
+    """
+    informatics type attach_compounds is constructed as a list of dict detailing
+    new compounds
+
+    Parameters
+    ----------
+    data:
```
yangchoo, line 38: nit: annotate with `dict`

autoprotocol/informatics.py
```diff
@@ -0,0 +1,119 @@
...
+        pass
+
+    @abstractmethod
+    def as_dict(self):
+        pass
+
+
+class AttachCompounds(Informatics):
+    """
+    informatics type attach_compounds is constructed as a list of dict detailing
```
yangchoo, line 33: nit: Capitalize?

autoprotocol/informatics.py
```diff
@@ -0,0 +1,119 @@
...
+        wells must be Well, list of Well, or WellGroup
+    ValueError
+        wells must be part of wells or containers instruction operates on
+    TypeError
+        compounds must be a list
+    TypeError
+        each element in compounds must be Compound type
+    """
+
+    def __init__(self, data: dict, all_wells):
```
yangchoo, line 62: :+1: Thanks for adding types. We should look to eventually type annotate everything and use a static-type checker like mypy.

autoprotocol/instruction.py
```diff
@@ -17,14 +17,21 @@ class Instruction(object):
 
     builders = InstructionBuilders()
 
-    def __init__(self, op, data):
+    def __init__(self, op, data, informatics=None):
         super(Instruction, self).__init__()
+        if informatics is None:
+            informatics = []
```
yangchoo, line 23: nit: default informatics directly to `[]` instead of None? Should be equivalent here.

autoprotocol/informatics.py
yangchoo: On more thought... it feels like we could simplify this signature? I.e. Specify `wells` and `compounds` directly, rather than using this `data` concept.

autoprotocol/informatics.py
```diff
@@ -0,0 +1,119 @@
...
+        # turn data into specific attributes for this type of informatics
+        if not isinstance(data, dict):
+            raise TypeError(f"informatics data: {data} must be provided in a dict.")
+
+        self.wells = data["wells"]
+        self.compounds = data["compounds"]
```
yangchoo, line 68: Yup, see above. This isn't as exhaustive check. If we simplify the signature, that will come naturally

autoprotocol/informatics.py
yangchoo: Hmm yea.. this is strange. Ideally, we should simplify this such that we don't have to pass `all_wells` directly and only validate this when we're serializing the Instruction.

autoprotocol/instruction.py
```diff
@@ -1401,6 +1414,7 @@ def __init__(self, resource_id, dests, measurement_mode="volume"):
                 "measurement_mode": measurement_mode,
                 "to": dests,
             },
+            informatics={"informatics": informatics},
```
yangchoo, line 1417: Ideally.. I think we should allow `Informatics`-class here, rather than forcing someone to use a dict here

yangchoo (commented):
Yea.. Let's start with figuring out what's the desired interface here.

There are two main patterns I can think of. 
1) Specify the informatics directly as part of the instruction creation. E.g. `p.provision(..., informatics=AttachCompounds(...))`. This is similar to what's proposed here, but instead of providing a dict, let's use the Informatics class(es) directly.
2) Attach the informatics after, similar to time_constraints. E.g. `p.provision(...); p.attach_informatics(p.get_instruction_index(), AttachCompounds(...))`

Let's assume we go with Option 1 for now. In that case, we'll probably want to do the check during the Instruction __init__ itself.

autoprotocol/instruction.py
aota001: updated :)

autoprotocol/protocol.py
```diff
@@ -5137,6 +5246,9 @@ def provision(self, resource_id, dests, amounts=None, volumes=None):
 
         measurement_mode = self._identify_provision_mode(provision_amounts)
 
+        if informatics is not None:
+            informatics = self.check_informatics(informatics, dests)
```
yangchoo, line 5250: hmm ideally we should do this as part of every Instruction, for every Informatics subclass? Thinking out loud, it'll be nice to define the checks just once and for each instruction with an Informatics, we'll automatically go through the defined checks.

I.e. have a `validate` abstractmethod defined for the base Informatics class, have each child class implement that method. During Instruction instantiation, we call the `validate` method for each attached Informatics object.

autoprotocol/informatics.py
aota001: `data` - I think we still want to have 'data' to be the common parameter across all types so they are modular, and can be deserialized automatically. I think we still want `data` to be the common parameter to do so? (unless we specify in Protocol for this type, these are arguments, but Protocol shouldn't have to know that?) The parameters within 'data' and validation could be different, and those can be defined in each class.
`all_well` - ya, thought it was weird, too... :p moved this check to outside.

autoprotocol/protocol.py
aota001: Ya... that was the idea initially to run validation in Instruction... but I couldn't find a way to run validation on wells (to make sure Informatics wells are included in the instruction wells) at that level...? So I moved it to Protocol...

aota001:
* Added `get_wells` method in Instruction class to fetch all wells associated with an instruction
* Revised to run all validations in Instruction and Informatics at the time of instantiation

autoprotocol/instruction.py
```diff
@@ -86,6 +100,119 @@ def filter_criteria(item):
...
+        for info in self.informatics:
+            info_keys = info.keys()
+            if set(info_keys) != set(valid_informatics_keys):
+                raise KeyError(
+                    f"informatics dict: {info} must have keys: {valid_informatics_keys}"
+                )
+            # validate wells only if `wells` param is present in informatics
+            try:
+                if info["data"]["wells"]:
+                    self._check_info_wells()
```
yangchoo, line 142: nit: this is a generalization that all future informatics instructions with wells will specify that in the `["data"]["wells"]` key. It is a fair assumption to make, but is implicit.
If we want to be stricter, we can instead scope this under the `attach_compounds` check, i.e. 
```
info_obj = AttachCompounds(info["data"])
self._check_info_wells()
```

autoprotocol/instruction.py
```diff
@@ -86,6 +100,119 @@ def filter_criteria(item):
...
+        the instruction.
+
+        Raises
+        -------
+        TypeError
+            wells are Well, list of Well or WellGroup
+        ValueError
+            Informatics wells are part of wells Instruction is operating on
+        """
+        for info in self.informatics:
```
yangchoo, line 166: I understand that we're guarding this caller-side (i.e. by making a check that `data/wells` is present before calling this function). However, would generally prefer to make such assumptions/checks directly into the function itself.

As currently defined, this will iterate over every `self.informatics` here and will result in a false negative if we add in a new Informatics instruction that doesn't have ["data"]["wells"]. 

Let's narrow the scope of this function and make it more explicit? E.g. make this function only contain the business logic of comparing two well groups, and shift the `get_wells` call to the caller-side above.

autoprotocol/instruction.py
```diff
@@ -1373,6 +1500,8 @@ class Provision(Instruction):
       Destination(s) for specified resource, together with volume information
     measurement_mode : str
       Measurement mode. Possible values are :py:class:`autoprotocol.constants.MEASUREMENT_MODES`
+    informatics : list(Informatics)
+      List of expected aliquot effects at the completion of this instruction
```
yangchoo, line 1504: It is unfortunate if we need to explicitly define this for every Instruction. As currently defined, this should be implicitly available to all Instructions. Its probably reasonable to move the documentation to the parent-class init method. I.e. in `Instruction` itself.

If we need/want the finer level of control over which Instruction should have informatics, would prefer a mix-in class instead. I.e. `InstructionWithInformatics` subclass. But I don't think we need this?

Otherwise... seems like we'll need to do this for all Instructions. If we want to be more future-proof, we could if generalize implicitly passing arguments to the parent class via keyword arguments. I doubt we'll add lots of top-level fields to Instruction though, so we may just need to go through and add `informatics` to every class :/ 
I.e.
```
def __init__(self, resource_id, dests, measurement_mode="volume", **kwargs)
...
super(Provision, self).__init__(
...
**kwargs
)
```

autoprotocol/informatics.py
yangchoo: More explicitly, instead of  `def __init__(self, data: dict):`, let's do `def __init__(self, wells, compounds):`

autoprotocol/protocol.py
```diff
@@ -5877,6 +5886,10 @@ def _refify(self, op_data):
             return self._refify(op_data._as_AST())
         elif isinstance(op_data, Ref):
             return op_data.opts
+        elif isinstance(op_data, Compound):
+            return op_data.InChI
+        elif isinstance(op_data, Informatics):
+            return op_data.as_dict()
```
yangchoo, line 5892: this doesn't work currently... we may need to do `self._refify(op_data.as_dict()`. 
We should add a test to _refify to verify this.

autoprotocol/protocol.py
```diff
@@ -5076,6 +5080,9 @@ def provision(self, resource_id, dests, amounts=None, volumes=None):
           one should be specified explicitly in a list matching the order of the
           specified destinations.
           Note:  Volumes and amounts arguments are mutually exclusive. Only one is required
+        informatics: list(dict)
```
yangchoo, line 5083: See above comment. This should be `list(Informatics)`. We could also additionally support `list(dict)` if that's desired.

autoprotocol/instruction.py
```diff
@@ -86,6 +100,119 @@ def filter_criteria(item):
             ]
         return data
 
+    def _set_informatics(self):
```
yangchoo, line 103: as per the previous review, I would prefer if we can pass an instantiated Informatics-class directly to the instruction-helper. I.e. `p.provision(..., informatics=[AttachCompounds(...)])`. 
Otherwise, if we do not want to end up manually adding defining that additional option for all Instructions, we can consider the `append` pattern, where you attach an Informatics to the Instruction in a separate call.  I.e. `p.provision(...); p.attach_informatics(AttachCompounds(...), instruction_index=p.get_instruction_index())`

Thinking more about it, I'm increasingly favoring the latter. Something that we can certainly get more user feedback on though.

yangchoo (changes requested):
Generally there. Biggest question remaining is the user-facing signature for specifying/attaching an Informatics to an existing Instruction.

autoprotocol/instruction.py
aota001: Ah. I guess so. good point.

autoprotocol/instruction.py
aota001: Ya, this was the main bottleneck that I had trouble figuring out...

autoprotocol/protocol.py
aota001: oops. sorry about that! will add test for this.

autoprotocol/protocol.py
aota001: Oh. I think I totally missed/misunderstood this. I thought we wanted to keep user input as a dict. my bad!

autoprotocol/instruction.py
aota001: Hmm. I don't mind `p.attach_informatics` route either, but we still need to associate `AttachCompounds` to a specific instruction. Should we have instruction index as a user input? Or assume `attach_informatics` to have effect on the previous instruction in the sequence of events?

aota001: oh, never mind. You did mention instruction_index. :p 
Ya, I initially thought of that (in https://github.com/strateos/autoprotocol_standard_changes/pull/16), but got turned down since it's harder to deserialize...

autoprotocol/instruction.py
aota001: Right. I changed this to `[]` then got warning saying `[]` is mutable. Read [this](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) and changed it back to `None`...

aota001:
@yangchoo 
* Updated AttachCompounds to take wells and compounds (instead of data)
* Update protocol to take in [Informatics] (instead of list(dict))
* revised checks in Instruction accordingly

autoprotocol/instruction.py
yangchoo: Ah, yup. That's right..

yangchoo (dismissed):
Was testing this more and encountered an odd bug.
```
p.provision("rs123", c1.well(1), volumes="10:microliter", informatics=AttachCompounds(c1.well(1), [cmpd]))
p.provision("rs123", c1.well(2), volumes="20:microliter")
```
This currently generates a valid Provision instruction, since the logic for appending modifies the `to` field in-place (L5185).

In the first-place, that's not a great pattern, but for backwards compatibility reasons, seems like we may need to add an additional check as late as possible, e.g. to the serialization (`_as_AST`) call.

Other than that, things look good for now. I think its still worth considering whether towards to `attach_informatics` style vs. adding directly to the protocol instruction helpers.

yangchoo (approved):
lgtm, thanks for all the changes!

PR #291, opened 2021-02-24
support: Update copyright year and active authors

yangchoo:
Update copyright year to 2021 and active authors list


aota001 (approved):
thanks for updating!

PR #292, opened 2021-02-24
feat: add informatics to LH

aota001:
@yangchoo 
Informatics need to be implemented in LH as well.
However, there's one issue... 
In LH, transfer form one source to multiple destinations (e.g. source=resourceA, dests = [aliquot1, aliquot2 aliquot3]) would instantiate 3 Instruction: Instruction("liquid_handle" for resourceA to aliquot1), Instruction("liquid_handle" for resourceA to aliquot2), Instruction("liquid_handle" for resourceA to aliquot3). This is a problem with how we are handling informatics, since informatics is provided at Protocol level (p.transfer), but the validations are happening at each Instruction... which means if Informatics has dest wells: [aliquot1, aliquot2, aliquot3], it fails (on informatics wells is part of Instruction wells) since at Instruction, there is one destination well at a time.
We could move validation back to the Protocol, or if we think LH is an edge case, we can have custom check just for this... thoughts???

### DIFF
--- a/autoprotocol/instruction.py
+++ b/autoprotocol/instruction.py
@@ -1849,16 +1849,19 @@ class LiquidHandle(Instruction):
         the liquid handling mode
     mode_params : dict, optional
         See Also :meth:`LiquidHandle.builders.instruction_mode_params`
+    informatics : list(Informatics), optional
+        List of Informatics describing the intended aliquot effects upon
+        completion of this instruction.
     """
 
     builders = LiquidHandleBuilders()
 
-    def __init__(self, locations, shape=None, mode=None, mode_params=None):
+    def __init__(self, locations, shape=None, mode=None, mode_params=None, informatics=None):
         data = {
             "locations": locations,
             "shape": shape,
             "mode": mode,
             "mode_params": mode_params,
         }
-
-        super(LiquidHandle, self).__init__(op="liquid_handle", data=data)
+        informatics = informatics
+        super(LiquidHandle, self).__init__(op="liquid_handle", data=data, informatics=informatics)
--- a/autoprotocol/instruction.py
+++ b/autoprotocol/instruction.py
@@ -1856,12 +1856,16 @@ class LiquidHandle(Instruction):
 
     builders = LiquidHandleBuilders()
 
-    def __init__(self, locations, shape=None, mode=None, mode_params=None, informatics=None):
+    def __init__(
+        self, locations, shape=None, mode=None, mode_params=None, informatics=None
+    ):
         data = {
             "locations": locations,
             "shape": shape,
             "mode": mode,
             "mode_params": mode_params,
         }
-        informatics = informatics
-        super(LiquidHandle, self).__init__(op="liquid_handle", data=data, informatics=informatics)
+
+        super(LiquidHandle, self).__init__(
+            op="liquid_handle", data=data, informatics=informatics
+        )
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -6792,6 +6792,59 @@ class Protocol(object):
             method = [method] * count
         method = [_validate_as_instance(_, Transfer) for _ in method]
 
+        # if informatics is provided for multiple wells, split Informatics for each destination well
+        # with the specified compounds.
+        if informatics is not None and len(informatics) > 0:
+            if len(informatics) == 1:
+                if isinstance(informatics[0], AttachCompounds):
+                    compounds = informatics[0].compounds
+                    wells = WellGroup(informatics[0].wells)
+                    if len(wells) == count and set(wells) == set(destination):
+                        info_compd = compounds * count
+                        informatics_list = []
+                        for well, compd in zip(destination, info_compd):
+                            if isinstance(informatics[0], AttachCompounds):
+                                if not isinstance(compd, list):
+                                    compd = [compd]
+                                if not isinstance(compd, list):
+                                    compd = [compd]
+                                informatics_list.append(AttachCompounds(well, compd))
+                    else:
+                        raise ValueError(f"Informatics wells: {wells} do not match wells used in Instruction.")
+                else:
+                    raise TypeError(f"Informatics:{informatics} is not available in this protocol.")
+            else:
+                wells_compounds_dict = {}
+                for info in informatics:
+                    if isinstance(info, AttachCompounds):
+                        wells_count = len(WellGroup(info.wells))
+                        compounds = info.compounds * wells_count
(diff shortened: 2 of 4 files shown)


autoprotocol/instruction.py
```diff
@@ -1849,16 +1849,19 @@ class LiquidHandle(Instruction):
...
+    def __init__(self, locations, shape=None, mode=None, mode_params=None, informatics=None):
         data = {
             "locations": locations,
             "shape": shape,
             "mode": mode,
             "mode_params": mode_params,
         }
-
-        super(LiquidHandle, self).__init__(op="liquid_handle", data=data)
+        informatics = informatics
```
yangchoo, line 1866: nit: this probably shouldn't be here?

yangchoo (approved):
Technically, informatics should be plumbed through for all Instruction-classes. We can punt on whether we should plumb it to all protocol helpers (i.e. `p.transfer`). 
I assume in this case that only `p.transfer` is required? There are other helpers which generate `LiquidHandle` instructions as well.

Review-wise, seems like you left a debug statement in, other than that, things look good.

aota001:
@yangchoo @dzleidig 
* added method to parse Informatics (based on the previous code which seems to only allow 1 destination for multiple sources, 1 source for multiple destinations, or same number of source and destination)
* Informatics wells can only include destination wells
* When there are multiple sources, Informatics is added only to the last Transfer Instruction (let me know if it should be handled differently?)
* When there are multiple destinations, len(informatics) must match len(destination)
* The order of Informatics wells can be different from destination wells as long as they contain the same wells, and the length match

Please let me know if I missed anything, feature should be changed, conceptually incorrect in any way, etc

autoprotocol/instruction.py
aota001: :p thanks!

autoprotocol/protocol.py
```diff
@@ -6745,10 +6748,109 @@ def location_helper(source, destination, volume, method, density):
...
+                Multiple instances of Informatics on a Well
+            ValueError
+                Parsed list of Informatics length should match the length of destination
+            """
+            if multi_src:
+                dest_count = 1
+            else:
+                dest_count = len(dest)
+
+            if len(informatics) == 1:
```
yangchoo, line 6786: ok, so this is the case where we try to intelligently parse potential intentions

test/protocol_test.py
```diff
@@ -3358,6 +3359,202 @@ def test_can_transfer_properties(self):
...
+                "10:microliter",
+                informatics=["foo"],
+            )
+        with pytest.raises(TypeError):
+            self.p.transfer(
+                resource.well(0).set_volume("40:microliter"),
+                test_wells[0],
+                "10:microliter",
+                informatics=["foo"],
+            )
```
yangchoo, line 3533: duplicate entry?

autoprotocol/protocol.py
```diff
@@ -6745,10 +6748,109 @@ def location_helper(source, destination, volume, method, density):
...
+                        wells_count = len(WellGroup(info.wells))
+                        compounds = info.compounds * wells_count
+                        for w, c in zip(WellGroup(info.wells).wells, compounds):
+                            if w not in wells_compounds_dict.keys():
+                                wells_compounds_dict[w] = c
+                            else:
+                                raise ValueError(
+                                    f"There are multiple instances of Informatics for well: {w}. The "
+                                    f"intent is ambiguous."
+                                )
```
yangchoo, line 6825: hmm... this may be overly strict (its okay on the first-pass, could leave a TODO if you don't want to tackle this now).

For this case
```
p.transfer(
a.well(0), [a.well(1), a.well(1)], "10:microliter", 
informatics=[AttachCompounds(a.well(1), [CompoundA]), AttachCompounds(a.well(1), [CompoundB])
)
```
It seems reasonable to expect this to be valid, i.e. I would expect there to be a Compound attached at the completion of each liquid handle instruction.

docs/changelog.rst
```diff
@@ -2,6 +2,7 @@
 Changelog
 =========
 
+* :feature:`292` Add informatics param to LiquidHandle
```
yangchoo, line 5: nit: Let's mention to `p.transfer` explicitly as well, since there are other helper methods for generating LiquidHandle instructions.

yangchoo (approved):
Looks almost there, bunch of duplicated code and a suggestion.

Other than that, things generally look good

autoprotocol/protocol.py
aota001: I'll add more documentations :)

autoprotocol/protocol.py
```diff
@@ -6745,10 +6748,109 @@ def location_helper(source, destination, volume, method, density):
...
+            ValueError
+                Parsed list of Informatics length should match the length of destination
+            """
+            if multi_src:
+                dest_count = 1
+            else:
+                dest_count = len(dest)
+
+            if len(informatics) == 1:
+                if isinstance(informatics[0], AttachCompounds):
```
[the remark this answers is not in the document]
aota001: Ah. Ok. I'll note that in comments as well.

autoprotocol/protocol.py
aota001: I see. That makes sense. I'll revise on this PR.

aota001:
@dzleidig Addressed comments from @yangchoo.
If you could take a quick look, that would be awesome. Thanks!

dzleidig (changes requested):
Overall looks good - implementation looks fine and test cases appear thorough. A few requests though:
- [x] Some top-level method documentation to guide the library-user through how we chose to handle some of the different scenarios, some limitations and opinions we built in. In-line comments are great, but we want to make it super clear for the library-user and avoid needing to have them walk through the code.
- [ ] An additional sad-path test case for attempts to apply an `informatic` to a source well, unless this is already covered in base class tests and is redundant.

yangchoo (approved):
Thanks for addressing the comments!

Hmm @dzleidig, right now this helper is pretty dumb, we're not accounting for any additional source/dest logic, e.g. we're not restricting AttachCompound to only be applicable to destination wells, or that Compounds have to be present in the source well. 
IMO, that's okay for now. I think the initial intent here is to provide an additional mechanism for the author to augment or extend informatics handling, and not necessarily to replace an entire informatics system by adding in more advanced logic. With that in mind, its fine to be less pedantic and strict on validating authors intent. 

With regards to public doc, agreed, let's add to the examples, i.e. ~L6700

yangchoo (approved):
Thanks for adding in the new documentation :)

PR #294, opened 2021-03-04
fix: bug on info wells should be a list

aota001:
AttachCompounds serializes wells into Well or list(Well). It should always be list(Well)

### DIFF
--- a/test/protocol_test.py
+++ b/test/protocol_test.py
@@ -2794,6 +2794,7 @@ class TestDyeTest(object):
         assert p1.instructions[3].informatics == []
         assert isinstance(p1.instructions[4].informatics[0], AttachCompounds)
         assert p1.instructions[4].informatics[0].compounds == [compd1]
+        assert p1.instructions[4].informatics[0].wells == [c1.well(1)]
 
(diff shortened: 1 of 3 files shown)


test/protocol_test.py
```diff
@@ -3446,8 +3448,8 @@ def test_transfer_informatics(self):
         assert new_instructions[3].informatics[0].compounds[0] == compd2
         wells = []
         for instr in new_instructions:
-            wells.append(instr.informatics[0].wells)
-        assert WellGroup(wells) == test_wells
+            wells.extend(instr.informatics[0].wells)
```
dzleidig, line 3451: Ah ok, so this was somehow working before because we were calling `append` which was appending a list to the list, which is wrong but will compile. 

`append` can take in a list or a single element. `extend` cannot though, so this test will break if `wells` is not a list. Should we modify this test to have a combination of instructions with multiple wells and instructions with single well?

aota001: Sounds good. Updated test :)

dzleidig (approved):

PR #297, opened 2021-03-16
feat: add SMILES to Compound

aota001:
* Change Compound to take in SMILES instead of InChI

### DIFF
--- a/autoprotocol/compound.py
+++ b/autoprotocol/compound.py
@@ -7,21 +7,53 @@ class Compound:
 
     Parameters
     ----------
-    SMILES : Simplified Molecular-Input Line-Entry System
+    format : Compound string input format
+    value : Compound string
 
     """
 
-    def __init__(self, SMILES):
-        if self.is_valid(SMILES):
-            self.SMILES = SMILES
+    def __init__(self, format, value):
+        compound_formats = ["Daylight Canonical SMILES", "InChI"]
+        if format in compound_formats:
+            self.format = format
         else:
-            raise CompoundError(f"{SMILES} is not a valid SMILES key")
+            raise CompoundError(f"{format} is not an acceptable Compound format.")
 
-    @staticmethod
-    def is_valid(SMILES):
-        # Any alphanumeric string with no space that does not start with 'InChI='
-        pattern = r"^(?!InChI=)\w\S+$"
-        return re.match(pattern, SMILES)
+        if self.is_valid(value):
+            self.value = value
+        else:
+            raise CompoundError(f"{value} is not a valid {self.format} value.")
+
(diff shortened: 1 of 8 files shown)


autoprotocol/compound.py
```diff
@@ -7,20 +7,21 @@ class Compound:
...
+            raise CompoundError(f"{SMILES} is not a valid SMILES key")
 
     @staticmethod
-    def is_valid(InChI):
-        pattern = r"^InChI\=1S?\/[^\s]+(\s|$)"
-        return re.match(pattern, InChI)
+    def is_valid(SMILES):
+        # Any alphanumeric string with no space that does not start with 'InChI='
+        pattern = r"^(?!InChI=)\w\S+$"
+        return re.match(pattern, SMILES)
```
dzleidig, line 24: Can we wrap this in a bool? `bool(re.match(pattern, SMILES))`

dzleidig (dismissed):
Left small nit - otherwise looks good. Thanks for the quick turnaround on this.

autoprotocol/compound.py
```diff
@@ -7,20 +7,58 @@ class Compound:
 
     Parameters
     ----------
-    InChI : Standard International Chemical Identifier
+    comp_format : Compound string input format
```
dzleidig, line 10: Let's just call this `format`, as it's within the context of a compound.

dzleidig (approved):
Left a small nit. We could also add some more assertions in the tests on the compound format.

autoprotocol/compound.py
```diff
@@ -7,20 +7,58 @@ class Compound:
...
+    def is_valid(self, compound):
+        inchi_pattern = r"^InChI\=1S?\/[^\s]+(\s|$)"
+        non_inchi_pattern = r"^(?!InChI=)\w\S+$"
+
+        if self.format == "InChI":
+            return bool(re.match(inchi_pattern, compound))
+        elif self.format == "Daylight Canonical SMILES":
+            return bool(
+                re.match(non_inchi_pattern, compound)
+                and not re.search("[~?!$%^&;'Jj]", compound)
```
dzleidig, line 36: Can you explain `re.search("[~?!$%^&;'Jj]"`? Perhaps add a comment and make `"[~?!$%^&;'Jj]"` a variable explaining what the regex is matching against.

If we're not 100% confident and clear on these regexs, we should remove this validation altogether.

aota001: Added comment. There are symbols that are not used in SMILES, but are used in other formats like SMART. Wanted to include these here in case we add other formats in the future.

dzleidig (approved):

PR #298, opened 2021-03-19
build: Release v7.7.0

aota001:
## Feature
#297 Add SMILES to Compound
#296 Remove low-evaporation cover type from FALCON96UBOTTOM


yangchoo (changes requested):
technically this should 7.7.0 if it includes a feature release. 
You may need to bump the bug to major in order for the documentation to properly render as well

yangchoo (approved):
lgtm, docs build fine too locally

PR #299, opened 2021-04-26
Add container attribute

aaronstrateos:
Added container attribute: is_reservoir
This boolean is evaluated in **utils.py** to bypass 384 well only stamping.

The intention is for stdev to overwrite this default property from false to true in internal protocols for clients when it makes sense. 

In the other branch called: **define-reservoir-from-attributes**, I define a reservoir from the container attributes, which is less robust and would likely cause errors for others consuming appy. Did it just for comparison.

### DIFF
--- a/autoprotocol/util.py
+++ b/autoprotocol/util.py
@@ -180,7 +180,8 @@ def _check_container_type_with_shape(container_type, shape):
                 f"1 or divisible by 2, but {shape} was specified."
             )
 
-    if shape["format"] == "SBS384" and container_wells < 384:
+# access reservoir attribute from predefined container type to determine if stampable
+    if shape["format"] == "SBS384" and container_wells < 384 and container_type.is_reservoir == False:
         raise ValueError(
             f"SBS384 transfers can only be executed in 384 well plates, but "
             f"container_type: {container_type} has {container_wells} wells."
--- a/autoprotocol/util.py
+++ b/autoprotocol/util.py
@@ -179,7 +179,7 @@ def _check_container_type_with_shape(container_type, shape):
                 f"containers must have row and column counts either equal to "
                 f"1 or divisible by 2, but {shape} was specified."
             )
-
+ 
 # access reservoir attribute from predefined container type to determine if stampable
     if shape["format"] == "SBS384" and container_wells < 384 and container_type.is_reservoir == False:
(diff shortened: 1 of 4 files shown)


autoprotocol/util.py
```diff
@@ -179,8 +179,9 @@ def _check_container_type_with_shape(container_type, shape):
...
                 f"1 or divisible by 2, but {shape} was specified."
             )
-
-    if shape["format"] == "SBS384" and container_wells < 384:
+ 
+# access reservoir attribute from predefined container type to determine if stampable
+    if shape["format"] == "SBS384" and container_wells < 384 and container_type.is_reservoir == False:
         raise ValueError(
             f"SBS384 transfers can only be executed in 384 well plates, but "
             f"container_type: {container_type} has {container_wells} wells."
```
aaronstrateos, line 187: There may be cases where a user wants to define a reservoir without requiring the shape and container wells to have specific constraints. That said, it might be better to do this instead:

`if container_type.is_reservoir == True:
     pass
    elif if shape["format"] == "SBS384" and container_wells < 384:
            raise ValueError(
                f"SBS384 transfers can only be executed in 384 well plates, but "
                f"container_type: {container_type} has {container_wells} wells." `

polarpine: What about checking on a container's `capabilities` property.  One challenge with `is_reservoir`, is that there are many types of reservoirs which may not be 384 well stamping compatible, eg. columnwise 12 well reservoirs, creating a mismatch of how to accurately describe a container while distinguishing valid instructions.  Additionally, and a little outside the intended scope of this task, is the case of a 1536-well plate.  A 1536-well plate can be used with a 384-well liquid handling head, similarly to how a 384-well plate can with a 96-pipette head.  This `is_reservoir` ContainerType property would not apply and an error would be thrown. Again, outside the scope of the immediate task, but worth keeping in mind.

Consider something along the lines of:
```
elif shape["format"] == "SBS384" and "sbs384_compatible" not in container_type.capabilities:
...
```
This would allow setting the known compatibility on a per container basis, which is helpful to differentiate a subclass of reservoirs.

aaronstrateos: Thanks yeah that makes sense. I submitted a pull request in a new branch and closed this one.

PR #300, opened 2021-04-27
enable 384 reservoir stamping if indicated in container_type.capabilities

aaronstrateos:
Instead of adding is_reservoir attribute, use the container_type.capabilities list to verify if plate is compatible with SBS384 transfers despite not having 384 wells.

* :release:`7.9.0 <2021-05-18>`
* :feature:`301` Allow reservoir stamping for plates with shape SBS384 provided that container has capability: `sbs384_compatible`. Capability added to container RESSW384LP.
* :support:`294` Add tests in util to check if container type is a compatible reservoir

**Please Land this PR after the stdv-665 port liquid_handle_dispense pr to respect versioning.**

[link to task](https://strateos.atlassian.net/secure/RapidBoard.jspa?rapidView=103&projectKey=STDV&modal=detail&selectedIssue=STDV-377&sprint=476)


aaronstrateos:
See former (now closed) pull request to compare different implementations:
[https://github.com/autoprotocol/autoprotocol-python/pull/299]

autoprotocol/util.py
```diff
@@ -180,10 +180,15 @@ def _check_container_type_with_shape(container_type, shape):
                 f"1 or divisible by 2, but {shape} was specified."
             )
 
-    if shape["format"] == "SBS384" and container_wells < 384:
+    if (
+        shape["format"] == "SBS384"
+        and container_wells < 384
+        and "sbs384_compatible" not in container_type.capabilities
```
EribertoLopez, line 186: container type capabilities are usually the names of the operations/instructions that can be performed on a container type. `sbs384_compatible` will never be found in the container_type.capabilities unfortunately.

aaronstrateos: @EribertoLopez In my other feature branch I added an is_reservoir attribute to the container_type dictionary with default of false to be set to true when writing client protocols. This approach above is per Vanessa's suggestion: [https://github.com/autoprotocol/autoprotocol-python/pull/299#discussion_r621251956](url)

In light of these two approaches, do you have an alternative idea?

aaronstrateos: And thanks for the feedback!

autoprotocol/container_type.py
```diff
@@ -6,7 +6,7 @@
     :license: BSD, see LICENSE for more details
 
 """
-
+# fmt: off
```
joshhacksthings, line 9: why are you turning off the Black formatter here? Don't like it? 🙈

test/util_test.py
```diff
@@ -1,7 +1,11 @@
+# fmt: off
+# pylint: disable=W0703
```
joshhacksthings, line 2: why are you turning black and pylint off?

test/util_test.py
```diff
@@ -25,3 +29,18 @@ def test_accepted_unit(self):
 
         with pytest.raises(TypeError):
             parse_unit("1:ul", ["second", "kg"])
+
+class TestUtil(object):
+    def test_compatible_reservoir_container(self):
```
joshhacksthings, line 34: I would like to see one more test that we _can_ stamp between 384-well plates and single-well reservoirs

joshhacksthings (changes requested):
left comments. lookin' good 😎

autoprotocol/container_type.py
aaronstrateos: Oh whoops, I'll fix that haha

test/util_test.py
aaronstrateos: 🙈

test/util_test.py
```diff
@@ -25,3 +28,35 @@ def test_accepted_unit(self):
...
+            source = p.ref("source", cont_type="res-sw384-lp", discard=True)
+            destination = p.ref("destination", cont_type="384-flat", discard=True)
+            p.transfer(
+                source=source.well(0),
+                destination=destination.well(0),
+                volume="10:microliter",
+                rows=16,
+                columns=24,
+            )
+        except Exception as exc:  # pylint: disable=W0703
```
joshhacksthings, line 61: I think I may have read this incorrectly above...

What is the _expected behavior_ here? Both methods in this test class `TestUtil` should **not** cause errors, right? I think wrapping these in try/except isn't the easiest approach.

Might I suggest something like:
1. One test that shows that we can transfer from 384-well plate -> 1-well reservoir
2. Another test showing that we still can't transfer from 384-well plate -> 96-well plate.... just to demonstrate that we're still hitting that expected Error because we're lacking `sbs384_ability` or whatever

joshhacksthings (changes requested):
i think i misunderstood your tests initially...

see my comment/question and lmk if what im asking makes sense

joshhacksthings (approved):
looks good! great test coverage 🚀

PR #302, opened 2021-06-07
PP-3846: Track Volume Loss in Evaporate Method

joshhacksthings:


cody-strateos (commented):
Reviewed for intent and it looks sane enough to me. I will leave approval for someone with more context around APPY.

test/protocol_test.py
```diff
@@ -3612,6 +3612,7 @@ def test_good_args(self):
             },
         )
         assert len(self.p.instructions) == 1
+        assert self.t1.well(0).volume == Unit(0, "uL")
```
aaronstrateos, line 3615: Maybe do a transfer to that well prior to evaporation method to make sure that the volume gets reset to zero.

aaronstrateos (dismissed):
Looks good. Made one comment about the evaporate test. Not essential, but probably a good idea.

aaronstrateos (dismissed):
LGTM!

julianicoleyang:
minor nit, but can you modify the title name to reference the jira ticket?

autoprotocol/protocol.py
```diff
@@ -7998,6 +8001,10 @@ def evaporate(self, ref, mode, duration, evaporator_temperature, mode_params=Non
                         f" {evaporator_temperature}"
                     )
 
+        # remove volume from all wells in evaporated container
+        for well in ref.all_wells():
+            well.set_volume(Unit(0, "uL"))
```
julianicoleyang, line 8006: Hmm. I will say that the one aspect that I'd be concerned about here is partial evaporation. Given that evaporation has params such as `duration` and `blow_rate`, one could also probably make an assumption that, if provided with certain parameters and original volume of the container, the evaporation may not necessarily go to completion.

Instead, this is zeroing out the volume irrespective of those parameters. Arguably, what we probably want here is `volume_to_subtract = blow_rate * duration` and then make it so that per well `volume = current_volume - volume_to_subtract` and ensure that volume cannot be less than zero here.

julianicoleyang (approved):
🚀

aaronstrateos (approved):

PR #308, opened 2021-08-27
PP-4666: LiquidHandleDispenseUpdate

EribertoLopez:
**Background**
Tempest is capable of conducting dispenses on all available chipsets on per pass of the instrument, meaning if there are 6 reagents to dispense, users can dispense all 6 of them from 6 chipsets on one warp. However, the driver right now is limited to using one dispense per one chipset per pass, which means users need 6 warps to dispense 6 reagents. It is a 6x increase in time operation. 

**Description**
The following changes allow users to to specify using multiple intake hoses to dispense to multiple destinations at once. To use multiple intake hoses from a single source, specify how many hoses to place in the source by specifying a List of Tuples of sources where the first element in the tuple is a Well and the second element is an int representing the number of intake hoses to place in the source specified.

```python
  Using multiple chips in a single source tube. The following example will
  dispense 100:microliters to 3 columns of the destination plate.

  .. code-block:: python

      from autoprotocol import Unit
      from autoprotocol import Protocol

      p = Protocol()
      source = p.ref("source", cont_type="micro-2.0", discard=True).well(0)
      destination_plate = p.ref("destination", cont_type="96-flat", discard=True)

      intake_hoses = 3
      source: List[Tuple[Well, int]] = [(source, intake_hoses)]

      destination = destination_plate.wells_from(0, 3)
      self.protocol.liquid_handle_dispense(
          source=source,
          destination=destination,
          volume='100:microliter',
      )
```

```
# Original liquid_handle_dispense instruction schema
    {
      "op": "liquid_handle",
      "locations": [
        {                                     # Aspirate from source
          "location": "src_tube_1/0", 
          "transports": [
            {"volume": "-710:microliter", "mode_params": {}}
          ]
        },
        {                                    # Dispense/Load Tempest Internal Chamber
          "location": "src_tube_1/0",
          "transports": [
            {"volume": "600:microliter", "mode_params": {}}
          ]
        },
        {                                  # Predispense into Tempest not Destination
          "transports": [
            {"volume": "10:microliter","mode_params": {}}
          ]
        },
        {                                 # Dispense into specified Destination
          "location": "flat/0",
          "transports": [
            {"volume": "100:microliter","mode_params": {}}
          ]
        }
      ],
      "shape": {"rows": 8, "columns": 1, "format": "SBS96"},
      "mode": "dispense"
    }

# Will generate this autoprotocol json
    {
      "op": "liquid_handle",
      "locations": [
        {                                     # Aspirate from source
          "location": "src_tube_1/0", 
          "transports": [
            {"volume": "-710:microliter", "mode_params": {}},
            {"volume": "-710:microliter", "mode_params": {}},
            {"volume": "-710:microliter", "mode_params": {}}
          ]
        },
        {                                    # Dispense/Load Tempest Internal Chamber
          "location": "src_tube_1/0",
          "transports": [
            {"volume": "600:microliter", "mode_params": {}},
            {"volume": "600:microliter", "mode_params": {}},
            {"volume": "600:microliter", "mode_params": {}}
          ]
        },
        {                                  # Predispense into Tempest not Destination
          "transports": [
            {"volume": "10:microliter", "mode_params": {}},
            {"volume": "10:microliter", "mode_params": {}},
            {"volume": "10:microliter", "mode_params": {}}
          ]
        },
        {                                 # Dispense into specified Destination
          "location": "flat/0",
          "transports": [
            {"volume": "100:microliter", "mode_params": {}}
          ]
        },
        {                                 # Dispense into specified Destination
          "location": "flat/1",
          "transports": [
            {"volume": "100:microliter", "mode_params": {}},
          ]
        },
        {                                 # Dispense into specified Destination
          "location": "flat/2",
          "transports": [
            {"volume": "100:microliter", "mode_params": {}}
          ]
        }

      ],
      "shape": {"rows": 8, "columns": 1, "format": "SBS96"},
      "mode": "dispense"
    }

### CODE REVIEW

File: autoprotocol/liquid_handle/dispense.py [hunk 1]
```diff
@@ -195,35 +236,41 @@ def default_volume_resolution(volume):
...
         Returns
         -------
         Unit
             prime volume
         """
-        return Unit(600, "uL")
+        if volume:
+            return parse_unit(volume, "ul")
+        else:
+            return Unit(600, "uL")
```
joshhacksthings, line 254: would it be reasonable to set the default value of `volume=Unit(600, "uL")`?

autoprotocol/liquid_handle/dispense.py
```diff
@@ -195,35 +236,41 @@ def default_volume_resolution(volume):
...
         Returns
         -------
         Unit
             predispense volume
         """
-        return Unit(10, "uL")
+        if volume:
+            return parse_unit(volume, "ul")
+        else:
+            return Unit(10, "uL")
```
joshhacksthings, line 273: could we set 10uL as default?

autoprotocol/protocol.py
```diff
@@ -1035,102 +1045,262 @@ def liquid_handle_dispense(
...
+        for dm, lc in zip(method, liquid):
+            # The index of the method and liquid classes should line up with
+            # source, destination, and volumes you wish to effect. The name of
+            # the liquid class will be applied to the transports of the mappings.
+            if not dm._liquid:
+                dm._liquid = lc
+
+        transport_locations = []
+        shape = LiquidHandle.builders.shape(rows, columns, None)
+        for i, multi_dispense_tup in enumerate(source):
```
joshhacksthings, line 1214: could you break out this `multi_dispense_tup` into something more explicit like:

`for i, (source_location, num_dispense_chips) in enumerate(source):`?

test/liquid_handle_dispense_test.py
```diff
@@ -14,16 +18,303 @@ def __init__(self):
...
+            sterile=False,
+            is_tube=False,
+            cover_types=["universal"],
+            seal_types=None,
+            capabilities=["cover"],
+            shortname="dummy",
+            col_count=12,
+            dead_volume_ul=Unit(15, "microliter"),
+            safe_min_volume_ul=Unit(30, "microliter"),
+        ),
```
joshhacksthings, line 45: why does ContainerType need to be passed here?

autoprotocol/protocol.py
```diff
@@ -1035,102 +1045,262 @@ def liquid_handle_dispense(
...
+                            for i in range(num_dispense_chips)
+                        ],
+                    )
+                )
+                for w in filled_destination_column_wells:
+                    w.add_volume(vol)
+
+            # Update source volume
+            if source_location.volume:
+                source_location.volume -= total_volume_dispensed
```
joshhacksthings, line 1293: caan you use `source_location.add_volume(- total_volume_dispensed)`?

joshhacksthings (dismissed):
lookin' good. i had some questions, but overall flow makes sense

autoprotocol/protocol.py
```diff
@@ -938,24 +940,33 @@ def liquid_handle_dispense(
...
-        source : Well
-            Well(s) to transfer liquid from.
-        destination : Well or WellGroup or list(Well)
-            Well(s) to transfer liquid to.
+        source : Well or list(Well) or list(tuple(Well, int))
+            Well(s) to transfer liquid from. If the source is given as a list
+            of (Well, int) tuples (ie: source=[(Well, 2), (Well, 5)])
+            the Well is the source well and the int is the number of dispense
+            chips to use from the specified source.
+        destination : Well or WellGroup or list(Well) or list(WellGroup)
```
joshhacksthings, line 948: in your example, this is a Container, not a wellgroup.

EribertoLopez:
Make sure that there are destination shape checks

autoprotocol/protocol.py
EribertoLopez: I may keep it the way it is to allow me to stub out the types, but will try your suggestion

autoprotocol/protocol.py
EribertoLopez: Yeah, I need to update the example sorry

test/liquid_handle_dispense_test.py
EribertoLopez: Don't quite remember why I put it there, will remove if possible

autoprotocol/liquid_handle/dispense.py
EribertoLopez: Yup, much cleaner!

autoprotocol/protocol.py
```diff
@@ -1154,14 +1154,14 @@ def format_source_well(well: Well, num_chips: int) -> Tuple[Well, int]:
...
-                    f"chips {number_of_chips} specified for the given source {(sw, number_of_chips)}"
-                )
+                for (sw, number_of_chips), destinations in zip(source, destination):
+                    # Check that the number of destinations match the number of chips specified for each
+                    # source destination mapping
+                    if len(destinations) != number_of_chips:
+                        raise ValueError(
+                            f"The number of destinations {len(destinations)} does not match the number of "
+                            f"chips {number_of_chips} specified for the given source {(sw, number_of_chips)}"
+                        )
```
ericschade, line 1164: we dont necessarily need the number of chips to match the number of destinations right?

we should be able to specify 

`sources = [(tube1.well(0), chips=4), (tube2.well(0), chips=1)]` with
 `destinations =[WellGroup(plate.A1), WellGroup(plate.A1)]`

but in this case, `len(src) = len(dest)`, but `len(destination[0]) != len(sources[0][1])`, which would trigger the above error message

ericschade: if i understand, we should be able to just remove this check completely since we dont really ever need the number of chips to match anything. we just have to make sure we dont have more chips than the total available - but you already check for that above

test/liquid_handle_dispense_test.py
```diff
@@ -92,22 +92,18 @@ def test_source_configuration(self):
...
                 does_not_raise(),
-                [WellGroup(self.flat.well(0)), WellGroup(self.flat.well(1))],
+                [[self.flat.well(0), self.flat.well(1)], self.flat.well(0)],
             ),
-            (
-                pytest.raises(ValueError),
-                [self.flat.well(0), self.flat.well(1), self.flat.well(0)],
-            ),
-            (does_not_raise(), [self.flat.well(0), self.flat.well(1)]),
+            (pytest.raises(ValueError), [self.flat.well(0), self.flat.well(1)]),
```
ericschade: related to other comment. maybe im forgetting a part of the conversation we had but i dont think we need the cihps to match the number of destination wells per source(because we dont know that each chip is going to dispense to exactly one location - the tempest software takes care of that decision making)

test/liquid_handle_dispense_test.py
```diff
@@ -203,7 +199,7 @@ def test_multi_chips_in_single_src_to_single_dest(
     ):
         source = [(self.src_tube_1, num_chips)]
         initial_src_volume = self.src_tube_1.volume
-        destination = [self.flat.well(0)]
+        destination = self.flat.wells_from(0, num_chips)
```
ericschade, line 202: wouldnt need to change this if im correct with the other comments

autoprotocol/protocol.py
```diff
@@ -1035,102 +1046,253 @@ def liquid_handle_dispense(
...
+            if len_vols != len_dests:
+                raise ValueError(
+                    f"When specifying more than one volume "
+                    f"the number of volumes must match the number of destination locations "
+                    f"num of volumes {len_vols} != num of destinations {len_dests}"
+                )
+            volume: List[Unit] = [
+                [parse_unit(v, "uL") for v in [vol] * len(destination[i])]
+                for i, vol in enumerate(volume)
+            ]
```
ericschade, line 1182: the way this is written currently will raise an error if the volumes are passed in as a list of 1 while there are more than one destination. 

not sure it makes a huge difference but passing in a list of 1 could be treated the same as a single volume.

autoprotocol/protocol.py
```diff
@@ -1035,102 +1046,253 @@ def liquid_handle_dispense(
...
+                    f"method list {method} did not contain all of type DispenseMethod"
+                )
+
+        # Set the DispenseMethod's liquid class
+        for dm, lc in zip(method, liquid):
+            # The index of the method and liquid classes should line up with
+            # source, destination, and volumes you wish to effect. The name of
+            # the liquid class will be applied to the transports of the mappings.
+            if not dm._liquid:
+                dm._liquid = lc
```
ericschade, line 1218: what are we protecting by writing it this way vs just setting `dm._liquid` without checking if there's already a liquid class?

autoprotocol/protocol.py
```diff
@@ -1035,102 +1046,278 @@ def liquid_handle_dispense(
                 volume=[Unit(_, "uL") for _ in range(1, 13)]
             )
 
+        Using multiple chips in a single source tube. The following example will
+        dispense 100:microliters to the first 3 columns of the destination plate.
```
GregoryEssertel, line 1050: I am still confused why the number of chips would have an impact on the destination wells.

Shouldn't that be totally independent? If we want to dispense to a single column and we request 3 chips, it shouldn't matter doesn't it? I think that become to linked to the device function.

autoprotocol/protocol.py
```diff
@@ -1035,102 +1046,278 @@ def liquid_handle_dispense(
...
-        if not isinstance(volume, list):
-            volume = [volume] * len(destination)
-        volume = [parse_unit(_, "uL") for _ in volume]
-
-        if not len(destination) == len(volume):
+        if remaining_num_chips_to_specify < 0:
             raise ValueError(
-                f"A different number of destinations: {destination} and volumes: {volume} were specified."
+                f"The number of chips configured for the "
+                f"instruction {default_max_num_dispense_chips - remaining_num_chips_to_specify}"
```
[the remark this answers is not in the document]
ericschade: i agree. im also thinking that we may not want a check for the maximum allowable number of chips to happen in autoprotocol. it should just give an error if the compiler cant find enough holders for all the specified chips. 

different tempests/non-tempest devices may have different numbers of chips(for ex. the one in tst3 had only 6 for months)

autoprotocol/protocol.py
ericschade: we're definitely on the same page, the docstring example just needs to be updated to reflect the code.

the new example should say `destination = [destionation.wells_from(0,3)]`

ericschade:
logging an interesting thing im seeing related to the failing test. In some cases its going to be useful to pass in a list of one list, which can be ugly/confusing(although i dont think we should remove it because it gives a cool functionality). the following works as expected and is elegant:
```
src = [(src_well_1, chips=2), (src_well_2, chips=3)]          # len(src) = 2
dest = [dest_well_1, dest_well_2]                                    # len(dest) = 2
```
the number of sources = number of destinations, so we know that src1 will be dispensed into dest1 and src2 will be dispensed into dest2.

 **howeverrrr** if we want a different specific behavior we can use the following:
```
src = [(src_well_1, chips=2), (src_well_2, chips=3)]          # len(src) = 2
dest = [[dest_well_1, dest_well_2]]                                  # len(dest) = 1
```
to have src1 dispense into dest1 + dest2.... AND ALSO have src2 dispense into dest1 + dest2

- because the length of dest is 1, there is logic which will copy that param to match the len(src), which ultimately results in the same transports as if we specified the following:
```
src = [(src_well_1, chips=2), (src_well_2, chips=3)]                               # len(src) = 2
dest = [[dest_well_1, dest_well_2], [dest_well_1, dest_well_2]]             # len(dest) = 2
```

autoprotocol/protocol.py
GregoryEssertel: I am still confused because now we specify 3 at two different places:

```
            intake_hoses = 3
            source: List[Tuple[Well, int]] = [(source, intake_hoses)]
            destination = [destination.wells_from(0, 3)]
```

What does the first 3 mean? and what the second 3 means?

Is the second 3 the number of column we want to dispense too? If yes can we add a variable name to make it more clear!

ericschade: the first 3 is the number of chips that should be fed into the source tube. the second 3 is an argument to the `.wells_from()` method which will return a wellgroup of `destination.well(0)`, `destination.well(1)` and `destination.well(2)`

the reason it's included in this example is because it makes more sense to have 3 chips dispense to 3 columns... but it could be any number

ericschade: yeah, i'll make it more clear as well though

GregoryEssertel: OK, that's what I guessed after, we could use 6 for the second to show it doesn't have to be equal!

ericschade: i changed it to 5... same thing though haha

ericschade:
@EribertoLopez on friday i fell into a rabbit hole trying to get that last test to pass from your last commit... which lead me to investigating the volume/destination configuration functions and getting lost in the logic haha(honestly dont even know if it was bugged - i was just very tired and stubborn) instead of trying to just change what was there i re-wrote them and all the tests pass now so 🤷 .


the destination function will turn every input into a List[WellGroup] of equal length to the source length

the volume function will turn every volume input into a List[List[Volume]] of equal length to the source length, where each item in the list is of the same length as the corresponding wg from destination.

i think this was the intent of the way you wrote it as well, but there was some corner case that was causing a test to fail and the recursion is complicated enough that it was easier for me to just write new ones rather than fix the existing

autoprotocol/protocol.py
GregoryEssertel: haha thanks, I think it is more clear now, and it makes sense

EribertoLopez:
> @EribertoLopez on friday i fell into a rabbit hole trying to get that last test to pass from your last commit... which lead me to investigating the volume/destination configuration functions and getting lost in the logic haha(honestly dont even know if it was bugged - i was just very tired and stubborn) instead of trying to just change what was there i re-wrote them and all the tests pass now so 🤷 .
> 
> the destination function will turn every input into a List[WellGroup] of equal length to the source length
> 
> the volume function will turn every volume input into a List[List[Volume]] of equal length to the source length, where each item in the list is of the same length as the corresponding wg from destination.
> 
> i think this was the intent of the way you wrote it as well, but there was some corner case that was causing a test to fail and the recursion is complicated enough that it was easier for me to just write new ones rather than fix the existing

@ericschade sounds good, thanks for the additions. Is there anything else that needs to change?

ericschade (approved):
ready to land this awesome new functionality! do we need to wait to merge on updates elsewhere in scle?

EribertoLopez:
> ready to land this awesome new functionality! do we need to wait to merge on updates elsewhere in scle?

Thanks! We will eventually, however it is a bit more involved in the backend so, the changes will be in dev branches until we test the changes end 2 end.

PR #316, opened 2021-10-18
Fix container mass extraction

joshhacksthings:
Extracting mass from containers was using the key `mass_mg` when we actually expect `mass` attribute (no underscore unit). This PR fixes that. Updated test.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :release:`unreleased <>`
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -2,6 +2,9 @@
 Changelog
 =========
 
+* :release:`7.9.6 <2021-10-18>`
+* :feature:`316` Add new container: `96-pcr-fs-clear`
```
EribertoLopez, line 6: @joshhacksthings update description of feature you're planning release with the change

PR #318, opened 2021-10-19
build: Bump pre-commit lib versions

yangchoo:
Bumps all pre-commit libraries except pylint to their latest versions.
There are no breaking changes introduced here.


EribertoLopez (dismissed):
LGTM! 🚀

PR #319, opened 2021-10-19
feat: Support Python 3.10

yangchoo:
Add Python 3.10 to the test suite as there are no obvious regressions.


yangchoo:
To merge only after https://github.com/autoprotocol/autoprotocol-python/pull/320 merges

PR #320, opened 2021-10-19
test: Bump pytest and pytest-cov versions

yangchoo:
Bump pytest and pytest-cov versions which are required for Python 3.10
support.


EribertoLopez (approved):
Thanks, Yang!

PR #332, opened 2021-11-11
Extract list of compounds, and associated metadata, when constructing aliquots with said data

joshhacksthings:
When we receive a JSON payload with the expected schema for compound declaration, we should associate these linked compounds to the `Well` they reside in.

This is the kind of structure we're expecting to receive, and we should only allow compounds with `id`, `molecularWeight`, and `smiles` keys to be associated, as we may need some of these keys in downstream applications. Any other keys to be considered should be included in a new ASC (autoprotocol standard change)
```
"compounds": [
    {.
        "id": "123",
        "molecularWeight": 100,
        "smiles": "CCCC",
    }
]
```

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -409,8 +409,10 @@ class Well(EntityPropertiesMixin):
                 raise RuntimeError(
                     "Compound information must include `id`, `molecularWeight`, and `smiles` keys."
                 )
-            # Transform string molecularWeight -> Unit molecularWeight
-            compound["molecularWeight"] = Unit(compound["molecularWeight"], "g/mol")
+            # Transform {"molecularWeight": float} -> {"molecular_weight": Unit}
+            compound["molecular_weight"] = Unit(
+                compound.pop("molecularWeight"), "g/mol"
+            )
 
         self.compounds = compounds
         return self
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -302,6 +302,7 @@ class Well(EntityPropertiesMixin):
         self.volume = None
         self.mass = None
         self.name = None
+        self.compounds = None
         self.properties = properties if isinstance(properties, dict) else dict()
         self.ctx_properties = self.fromDict(ctx_properties)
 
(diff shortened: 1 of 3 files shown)


xiliuStrateos:
Just a minor comment, the `molecularWeight` is not a python style. How/where is the schema defined?

aaronstrateos (approved):

EribertoLopez (approved):

test/manifest_test.py
```diff
@@ -753,3 +753,39 @@ def test_invalid_compound(self):
...
+                                ],
+                            }
+                        },
+                    }
+                },
+                "parameters": {"cont": "echo_plate"},
+            },
+        )
+        expected_compounds_list = [
+            {"id": "123", "molecularWeight": Unit(100, "g/mol"), "smiles": "CCCC"}
```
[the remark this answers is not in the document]
joshhacksthings: yeah so this is the "expected" value that should be returned.

what happens is:
1. web sends us the JSON, where molecularWeight type is float
2. here in APPY, we convert it to a Unit, and store it on the `compounds` attribute of a `Well`
3. So, when we call `well.compounds`, we're expecting this dict to be transformed to have the float value be a Unit

joshhacksthings:
> Just a minor comment, the `molecularWeight` is not a python style. How/where is the schema defined?

@xiliuStrateos I agree with you here. This schema is defined form our webapp, let me convert it to `molecular_weight`, on extraction

autoprotocol/container.py
```diff
@@ -375,6 +376,47 @@ def set_volume(self, vol):
...
+            Well with associated compounds
+
+        Raises
+        ------
+        TypeError
+            Incorrect input-type given
+        RuntimeError
+            Compound information does not contain required keys
+        """
+        expected_keys = {"id", "molecularWeight", "smiles"}
```
xiliuStrateos, line 400: Update the key as well.

xiliuStrateos (changes requested):

autoprotocol/container.py
joshhacksthings: So idk if I should do that here... we're getting these keys from web, then this function transforms `molecularWeight` --> `molecular_weight`

xiliuStrateos: Gotcha. All good then.

xiliuStrateos (approved):

PR #334, opened 2021-11-18
feat: add concentration & solubility flag in well (aliquot) compounds

MNikhil0210:
Add concentration & solubility_flag in aliquot compounds.

[Corresponding Web PR](https://github.com/strateos/web/pull/2236)

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -397,7 +397,7 @@ class Well(EntityPropertiesMixin):
         RuntimeError
             Compound information does not contain required keys
         """
-        expected_keys = {"id", "molecularWeight", "smiles"}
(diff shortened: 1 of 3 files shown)


EribertoLopez (dismissed):
So far change look good to me and in line with `/web` updates. Can we add the new fields to [this](https://github.com/autoprotocol/autoprotocol-python/blob/master/test/manifest_test.py#L757) test or make a separate one, thanks!

joshhacksthings (changes requested):
can you add a line to the changelog?
additionally, you need to update the function in `harness.py` that extracts this payload from JSON

EribertoLopez (approved):

PR #337, opened 2021-12-04
Update `set_compounds` on Well to allow compounds without all properties

polarpine:
Compounds should be set to an aliquot (Well) even if all information is not present.  This change introduces more flexibility to set parameters associated with compounds to a given Well.


ericschade (dismissed):
setting compounds won't cause an error if missing some properties. lgtm

EribertoLopez (approved):

ericschade (approved):

PR #341, opened 2021-12-07
fix: `millimol` -> `millimole`

joshhacksthings:
Fix a typo we introduced when creating a concentration unit


joshhacksthings:
Closing as our Units lib handles this fine

PR #343, opened 2022-01-07
Stdv 1126 corning 4513 container type

ericschade:
Adding the Corning 4513 container type to APPy.


ericschade:
cleaned up commit history with force push

autoprotocol/container_type.py
```diff
@@ -1071,6 +1071,42 @@ def row_count(self):
     cat_no="3657",
 )
 
+#:
+ROUND384LV = ContainerType(
+    name="384-well round-bottom low-volume non-binding-surface Corning-4513 plate",
+    well_count=384,
+    well_depth_mm=6.58,
+    well_volume_ul=Unit(20.0, "microliter"),
+    well_coating="Corning NBS(nonbinding surface)",
```
ericschade, line 1080: im pretty sure this can be anything - i just went to the vendor site and transcribed from there

autoprotocol/container_type.py
```diff
@@ -1373,6 +1409,7 @@ def row_count(self):
     "384-v-clear-clear": V384CLEAR,
     "384-round-clear-clear": ROUND384CLEAR,
     "res-sw384-lp": RESSW384LP,
+    "384-Corning-4513-round-lv": ROUND384LV,
```
joshhacksthings, line 1412: no capitals in dash-case pls!

autoprotocol/container_type.py
```diff
@@ -1071,6 +1071,42 @@ def row_count(self):
...
+        "cover",
+        "uncover",
+        "echo_dest",
+        "stamp",
+        "transfer",
+        "provision",
+        "pipette",
+        "envision",
+    ],
+    shortname="384-Corning-4513-round-lv",
```
joshhacksthings, line 1102: update here once uncapitalized below

joshhacksthings (changes requested):
almost there:

- [x] remove capitalized letter in dash-case shortname
- [ ] Check for other existing NBS coatings, would suggest conforming to any existing string that has NBS in it already

ericschade:
> almost there:
> 
> * [x]  remove capitalized letter in dash-case shortname
> * [ ]  Check for other existing NBS coatings, would suggest conforming to any existing string that has NBS in it already

I don't see any other examples of NBS in the description of plates. do you think that it should be described differently than it is here?

joshhacksthings (approved):
lgtm

PR #346, opened 2022-05-19
Release 7.14.3

polarpine:


ericschade (approved):
new container type for sample management 🤙

joshhacksthings (approved):
lgtm! i would personally argue you could get away with a minor bump here --> 7.15.0, given that the new container type is a new feature, but i also understand wanting to keep it a patch as it's ultimately a small addition in the grand scheme of things

PR #350, opened 2022-10-20
pp-10468: remove pump_ovverides

sabrinamohabbat:
https://strateos.atlassian.net/browse/PP-10468

remove pump_override_volume so that it defaults to none


aaronstrateos (commented):
Can you post a test run to this? Downside of this approach is that it remove's pump-override vol so that it is always None, rather than allowing us to set it at the top level of a LHD function call.

sabrinamohabbat:
> Can you post a test run to this? Downside of this approach is that it remove's pump-override vol so that it is always None, rather than allowing us to set it at the top level of a LHD function call.

https://secure.strateos.com/l2s2dev/p1e2ac4zc5wkqp/runs/r1hkxd5xxfc3w7/instructions Test run here

aaronstrateos:
pump override vol still showing up in ap.json
Screen Shot 2022-10-20 at 5 14 48 PM

EribertoLopez:
> pump override vol still showing up in ap.json Screen Shot 2022-10-20 at 5 14 48 PM

The pump override in the example run only shows up once due to the custom LiquidClass being passed in the .transfer(). We still want to allow the developer to configure this. However, you don't see the additional pump_overrides that these changes remove from the liquid_handle instruction.

EribertoLopez (approved):

PR #352, opened 2022-10-24
Enhanced error message(s) when Well.set_volume() exceeds maximum allowed

valtriadi:
When Well.set_volume() exceeds maximum allowed, the error message returned should include Container name, ID (if available) and the Well index number.

** STDV Internal Only **
(The comment below may be removed before merge to origin repo due to the internal STDV-only info)
Ticket: https://strateos.atlassian.net/browse/STDV-823
Test: tox script validation.

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -255,11 +255,11 @@ class Well(EntityPropertiesMixin):
         if v > max_vol:
             containerIdInfo = ""
             if self.container.id is not None and len(self.container.id) > 0:
-                containerIdInfo = "with container id [" + self.container.id + "]"
+                containerIdInfo = " with container id " + self.container.id
 
             raise ValueError(
-                f"Failure at container [{self.container.name}] {containerIdInfo} "
-                f"when setting the volume for well [{self.index}]: "
+                f"Failure at container '{self.container.name}'{containerIdInfo} "
+                f"when setting the volume for well {self.index}: "
                 f"Theoretical volume [{v}] to be set exceeds maximum well "
                 f"volume [{max_vol}]."
             )
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -255,7 +255,7 @@ class Well(EntityPropertiesMixin):
         if v > max_vol:
             containerIdInfo = ""
             if self.container.id:
-                containerIdInfo = " with container id " + self.container.id
+                containerIdInfo = f" with container ID: {self.container.id}"
 
             raise ValueError(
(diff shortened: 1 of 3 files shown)


autoprotocol/container.py
```diff
@@ -253,10 +253,17 @@ def set_volume(self, vol):
         v = Unit(vol)
         max_vol = self.container.container_type.true_max_vol_ul
         if v > max_vol:
+            containerIdInfo = ""
+            if self.container.id is not None and len(self.container.id) > 0:
```
joshhacksthings, line 257: I think we can simplify this to:

```suggestion
            containerIdInfo = ""
            if self.container.id:
```

autoprotocol/container.py
```diff
@@ -253,10 +253,17 @@ def set_volume(self, vol):
...
+                containerIdInfo = " with container id " + self.container.id
```
joshhacksthings, line 258: I think we should capitalize ID here, since it'll be in a human-read string, also let's f-string it to get some helpful single quotes ie

```suggestion
                containerIdInfo = f" with container ID: {self.container.id}"
```

autoprotocol/container.py
```diff
@@ -253,10 +253,17 @@ def set_volume(self, vol):
...
+
             raise ValueError(
-                f"Theoretical volume {v} to be set exceeds maximum well "
-                f"volume {max_vol}."
+                f"Failure at container '{self.container.name}'{containerIdInfo} "
+                f"when setting the volume for well {self.index}: "
+                f"Theoretical volume [{v}] to be set exceeds maximum well "
+                f"volume [{max_vol}]."
```
joshhacksthings, line 264: this is close... but maybe try...

```suggestion
                f"Theoretical volume [{v}] to be set exceeds maximum well "
                f"volume [{max_vol}] for container '{self.container.name}'{containerIdInfo} "
                f"when setting the volume for well {self.index}."
```

joshhacksthings (dismissed):
this looks good, just made some style/wording suggestions. lmk if you have questions! otherwise, you're clear to merge once you've made these changes 😄

valtriadi:
@joshhacksthings ~ Thank you for the review + feedback! I applied all of your code change suggestions, and the tox validations are all passed. But somehow weird things happened: Not only I lost your "approval" but also GIT said I "_dismissed your stale review_" ... I did NOT dismiss anything, so not sure what it meant. I re-request your review. Sorry not sure what happened here, but the feedback changes were committed.

joshhacksthings (approved):
lgtm!! 🚀

PR #358, opened 2022-11-06
[STDV-1819]: Dataclasses enhancements to support objects schema-ready and de/serialization.

valtriadi:
Included: Container, Unit, and Well

***
**FAILURES**
Current tox failures (tests, lint and docs): All related to Python 3.6 (default system environment) unable to recognize module: dataclasses.

**1. autoprotocol-python\.tox\py36**
```
autoprotocol\container.py:12
    from dataclasses import dataclass, field
    ERROR:   ModuleNotFoundError: No module named 'dataclasses'
```

**2. lint run-test: pre-commit run --all-files**
```
Pylint ....... Failed
autoprotocol\unit.py:10:0: E0401: Unable to import 'dataclasses' (import-error)
autoprotocol\container.py:12:0: E0401: Unable to import 'dataclasses' (import-error)
```

**3. docs run-test: sphinx-build -W -b html** 
```
autodoc: failed to import class 'unit.Unit' from module 'autoprotocol'; the following exception was raised:
No module named 'dataclasses'
```

***
**INSIGHTS**
It appears only py37 and beyond supports dataclasses module. Weird that online chatters were discussing as-if this module was available since py3.6 (it is not). There was suggested solution to include pulling/install dataclasses directly when using py36.

REF:

- Python dataclasses: https://docs.python.org/3/library/dataclasses.html
- PEP 557: https://peps.python.org/pep-0557/
- Suggested solution: https://github.com/spulec/moto/issues/4301


EribertoLopez:
@yangchoo @joshhacksthings can we depreciate python 3.6 to allow the use of @dataclass decorator? I believe Igor runs python 3.7.6

yangchoo:
Yea, I think we can, but if we do, it has to be a major version bump. Let's handle that in a separate PR?

EribertoLopez:
Must land after https://github.com/autoprotocol/autoprotocol-python/pull/359

autoprotocol/unit.py
```diff
@@ -183,6 +185,9 @@ class Unit(_Quantity):
 
     """
 
+    value: Any
+    units: Optional[Any] = None
```
EribertoLopez, line 189:
```suggestion
    value: Union[int, float, str]
    units: Optional[str] = None
```

test/builder_test.py
```diff
@@ -13,7 +13,7 @@
     Spectrophotometry,
     Thermocycle,
 )
-from autoprotocol.unit import UnitError
+from autoprotocol.unit import Unit, UnitError
```
EribertoLopez, line 16: Doesn't seem like you are using the Unit. Remove unused import

EribertoLopez (approved):
LGTM! Thanks, @valtriadi 🚀

PR #365, opened 2022-12-05
feat: PP-11306 - Adds multiple options for autopick allowing for multiple groups per instruction

AlexanderNull:
Enables providing multiple groups of picks per single autopick instruction

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -5083,9 +5083,7 @@ class Protocol(object):
         groups = []
         for group in pick_groups:
             groups.append(
-                self.__process_pick_group(
-                    group.source, group.destination, group.min_abort
-                )
+                self.__process_pick_group(group)
             )
 
         # Current device requirement is that all autopick group sources are from the same container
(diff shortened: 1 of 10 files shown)


autoprotocol/types/protocol.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
```
ericschade, line 1: this is the additional import you mentioned which is the tiebreaker in terms of which option you prefer?

ericschade: nvm. when you say consumers youre talking about ap writers, yeah? the import would be `from protocol.types import pickparams`

ericschade:
im also pretty neutral between the two options but probably lean slightly towards the param class because i like the idea of being able to 

```
myparams = AutopickParams(source = s, dest = d, min_abort = m_a)
```
or 
```
myparams = AutopickParams(dest = d, source = s, min_abort = m_a)
```
and not have to remember that tuple[0] is the source while tuple[1] is the dest(either order would pass the type hint since both are of type `WellParam`)

autoprotocol/protocol.py
```diff
@@ -5081,6 +5083,97 @@ def autopick(
             Source wells are not all from the same container
 
         """
+        groups = []
+        for group in pick_groups:
+            sources, dests, min_abort = pick_groups
```
AlexanderNull, line 5088: ... = group

test/protocol_test.py
```diff
@@ -1691,14 +1732,46 @@ def test_autopick(self):
...
+            "agar_plate",
+            {"reserve": "ki17reefwqq3sq", "discard": True},
+            Container(None, p.container_type("6-flat"), name="agar_plate"),
         )
 
+        agar_plate = Container(None, p.container_type("6-flat"), name="agar_plate")
+        source1 = [agar_plate.well(0), agar_plate.well(1)]
+        dest1 = [dest_plate.well(1)] * 4
+        pick_group = AutopickGroupClass(source1, dest1)
+        pick_group = ([agar_plate.well(0), agar_plate.well(1)], [dest_plate.well(1)] * 4, 0)
```
AlexanderNull, line 1771: remove second duplicate line

autoprotocol/types/protocol.py
AlexanderNull: an example of it is down in the test that uses autopick2

autoprotocol/protocol.py
```diff
@@ -5081,6 +5082,27 @@ def autopick(
             Source wells are not all from the same container
 
         """
+        groups = []
```
EribertoLopez, line 5085: It may be good to have a check that checks that the group is an instance of the AutopickGroup class.

aaronkarlsberg1 (dismissed):
lgtm, stronger typing for the win! Echoing Eddie's suggestion of runtime check that the group is of type AutopickGroup.

EribertoLopez (dismissed):

EribertoLopez (approved):

PR #369, opened 2022-12-13
stdv-1949-ap-python-typing

sabrinamohabbat:
This PR types all dictionaries in arguments either by making a dataclass or by defining the key value (ie: Dict[str, str] which means the key and value are string types). 

[jira ticket](https://strateos.atlassian.net/browse/STDV-1949)

[STDV-1949]: https://strateos.atlassian.net/browse/STDV-1949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


autoprotocol/builders.py
```diff
@@ -2662,14 +2761,18 @@ def channel(
         if trigger_threshold is not None and not isinstance(trigger_threshold, int):
             raise TypeError("trigger_threshold must be of type int.")
 
-        trigger_modes = ("and", "or")
+        trigger_modes = ("and_", "or_")
```
EribertoLopez, line 2764: TODO: revisit this

autoprotocol/builders.py
```diff
@@ -2015,14 +2023,21 @@ def incubate_params(
...
+
 class EvaporateBuilders(InstructionBuilders):
     """
     Helpers for building Evaporate instructions
     """
 
     def __init__(self):
         super(EvaporateBuilders, self).__init__()
-        self.valid_modes = ["rotate", "centrifuge", "vortex", "blowdown"]
+        self.valid_modes = ["rotary", "centrifugal", "vortex", "blowdown"]
```
EribertoLopez, line 2040: # TODO: Revisit this and find which valid_modes are true

autoprotocol/builders.py
```diff
@@ -2679,10 +2782,15 @@ def channel(
             "detector_gain": detector_gain,
             "measurements": measurements,
             "trigger_threshold": trigger_threshold,
-            "trigger_logic": trigger_logic,
+            "trigger_logic": trigger_logic[-1:],
```
EribertoLopez, line 2785: TODO: I don't think we need to slice the string, revert

autoprotocol/builders.py
sabrinamohabbat: I believe the valid modes are rotate and centrifuge, here is the [commit](https://github.com/strateos/autoprotocol-scala/blob/ce7db79a4a80ee686f301932e805d8d9a80af997/src/autoprotocol/src/main/scala/org/autoprotocol/instructions/Evaporate.scala) that I found that makes me think this.

autoprotocol/protocol.py
```diff
@@ -282,36 +261,36 @@ def ref(
...
-                opts["new"] = cont_type.shortname
+                opts.new = cont_type.shortname
         except ValueError as e:
             raise RuntimeError(
                 f"{cont_type} is not a recognized container type."
             ) from e
 
         if storage:
-            opts["store"] = {"where": storage}
+            opts.store = StorageLocation(**{"where": storage})
```
valtriadi, line 279: Storage as [Location enum](https://github.com/autoprotocol/autoprotocol-python/blob/stdv-1949-update-ap-python-protocol-instruction-arguments/autoprotocol/types/ref.py#L10-L16) inside types/ref.py > VERSUS Storage as [regular String](https://github.com/autoprotocol/autoprotocol-python/blob/stdv-1949-update-ap-python-protocol-instruction-arguments/autoprotocol/protocol.py#L171) inside  protocol.py > ref() : 

I'm curious if this difference between the input params could cause breakage because currently the protocol.py > ref() function allows user to pass in any String value for storage param; but this line of code for RefOpts datatype expect specificaly for Location enum ? 

While all of ApPy tests passed given this codebase condition, I wonder if we can add new test case for protocol.ref() where the Storage input would be any String NOT existing in Location enum, and anticipate error/success?

valtriadi (approved):
LGTM 👍 

Nit: I was curious about a test case/scenario mentioned above.

autoprotocol/protocol.py
valtriadi: NOTE: @sabrinamohabbat  wrote the test in question where we expected test to fail given invalid storage value = "hot" (not applicable in Location enum). But the test unexpectedly pass (success). Abandon the test request for the moment; and I will need separate study why Python would enable not valid String in the place of enum in the method signature.

valtriadi (approved):
Thank you @sabrinamohabbat 👍

PR #372, opened 2022-12-15
Added property propagation to acoustic transfer

jguiel:


autoprotocol/protocol.py
```diff
@@ -1679,6 +1679,8 @@ def acoustic_transfer(
                 s.volume -= v
             if v > Unit(0, "microliter"):
                 transfers.append(xfer)
+            if self.propagate_properties:
+                d.add_properties(s.properties)
```
cj-bonilla, line 1683: `add_properties()` works, but I think we want to use `set_properties()` instead? In case there are duplicate keys.

test/protocol_test.py
```diff
@@ -1377,6 +1377,20 @@ def test_droplet_size(self, dummy_protocol):
...
+        p.propagate_properties = True
+        echo = p.ref("echo", None, "384-echo", discard=True)
+        dest = p.ref("dest", None, "384-echo", discard=True)
+        echo.well(0).add_properties({"test_well_property": True})
+        p.acoustic_transfer(
+            echo.well(0).set_volume("2:microliter"),
+            dest.well(0),
+            "50:nanoliter"
+        )
+        echo_prop = echo.well(0).properties["test_well_property"]
```
cj-bonilla, line 1391: I didn't know you could write it like this! I thought it had to be `well.properties.get["key"]`

cj-bonilla (approved):
lgtm!

test/protocol_test.py
joshhacksthings: the `properties` attribute is a dictionary! you can use either `.get("key_name")` or `["key_name"]`

there's some implicit value over `get` as it returns `None` instead of raising a KeyError, but this value is dependent on your use case

autoprotocol/protocol.py
joshhacksthings: ^^ I think I agree with @cj-bonilla here! this could produce a UserWarning in the stdout, which is probably fine, if not helpful to user

joshhacksthings (changes requested):

test/protocol_test.py
jguiel: and I didnt realize you can run .get to avoid keyerrors 😂 we're learning _together_ @cj-bonilla

test/protocol_test.py
```diff
@@ -1377,6 +1377,22 @@ def test_droplet_size(self, dummy_protocol):
...
+    def test_propagate_properties(self, dummy_protocol):
+        p = dummy_protocol
+        p.propagate_properties = True
+        echo = p.ref("echo", None, "384-echo", discard=True)
+        dest = p.ref("dest", None, "384-echo", discard=True)
+        echo.well(0).set_properties({"test_well_property": True})
+        dest.well(0).set_properties({"test_well_property": False})
+        p.acoustic_transfer(
+            echo.well(0).set_volume("2:microliter"), dest.wells(0, 1), "50:nanoliter"
+        )
```
joshhacksthings, line 1389: super nit, but can you call `echo.well(0).set_volume("2:microliter")` above this input? then have this source arg just be `echo.well(0)` 

i think it'd be more readable, ie

```suggestion
        dest.well(0).set_properties({"test_well_property": False})
        echo.well(0).set_volume("2:microliter")

        p.acoustic_transfer(
            echo.well(0), dest.wells(0, 1), "50:nanoliter"
        )
```

joshhacksthings (approved):
looks great! i just had 2 things -

1. Move the echo set volume call for readability
2. Add a line in the changelog about your new feature!

joshhacksthings (dismissed):
conditionally approved, as you still need to Add a line in the changelog about your new feature

jguiel:
@cj-bonilla mentioned changelog should be updated in separate PR and any attempt to add `Unreleased` in some form or another into the changelog would fail `tox -e docs`, so I reverted changes. Possible solution could be combining our features into one version

docs/changelog.rst
```diff
@@ -1,6 +1,9 @@
 =========
 Changelog
 =========
+
+* :feature:`36X` Acoustic transfer propagates well properties
```
joshhacksthings, line 5:
```suggestion
* :feature:`372` Acoustic transfer propagates well properties
```

joshhacksthings (approved):

joshhacksthings (approved):

PR #378, opened 2023-02-01
SCIENG-154: Update 96-deep-kf max well volume in AP-py

cj-bonilla:
Updated max well volume of 96-deep-kf to 2mL, as specified in dimensions sheet

Originally, 1mL, but true max vol is 2mL. Please see ticket [SCIENG-154](https://strateos.atlassian.net/browse/SCIENG-154) for more information

[SCIENG-154]: https://strateos.atlassian.net/browse/SCIENG-154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cj-bonilla:
Converting to draft for now until I can figure out what the true safe max vol is. 

_Question to answer_: Is 1mL the max we can have when accommodating for the combi tips?

autoprotocol/instruction.py
```diff
@@ -14,7 +14,7 @@
 from .informatics import AttachCompounds, Informatics
 
 
-class Instruction(object):
+class Instruction:
```
sal-strateos, line 17: Any reason for this change. This could open up a can of worms depending on how this class is used especially if we have multiple inheritance downstream. Some things to keep in mind for new style vs. old style classes

image

If we have tests around this then I think the change should be fine.

autoprotocol/instruction.py
```diff
@@ -57,9 +57,11 @@ def _as_AST(self):
         """
         # informatics is not serialized if empty.
         if self.informatics:
-            return dict(op=self.op, **self.data, informatics=self.informatics)
+            return_dict = dict(op=self.op, **self.data, informatics=self.informatics)
```
sal-strateos, line 60: nit: I think it was fine as is since dict() is called along with the return. You could add typehints to the class to see a summary output when referencing

autoprotocol/instruction.py
```diff
@@ -250,7 +257,22 @@ def __init__(self, groups, magnetic_head):
                 f"container_types: {self.heads[magnetic_head]} for head_type: "
                 f"{magnetic_head}"
             )
-
+        non_valid_well_volumes = [
+            w
+            for _ in containers
```
sal-strateos, line 262: `_` is typically used when the return will not be used. In this case it is valuable to know what `_` represents

autoprotocol/instruction.py
```diff
@@ -250,7 +257,22 @@ def __init__(self, groups, magnetic_head):
                 f"container_types: {self.heads[magnetic_head]} for head_type: "
                 f"{magnetic_head}"
             )
-
+        non_valid_well_volumes = [
```
sal-strateos, line 260: Does this contain well volumes or well objects with with volumes above some threshold?

sal-strateos: `wells_with_invalid_volume` is what I think this should be

sal-strateos, line 260: If you want a bit of separation this init validation could be a classmethod that you call and has access to the class var `working_vols`

sal-strateos (approved):
Mostly looks good to me with some minor feedback. I do have one concern re. removal of inherited object for the Instruction class

autoprotocol/instruction.py
```diff
@@ -174,14 +177,14 @@ def get_wells(self, op_data):
             all source and destination wells for instructions such as `liquid_handle`.
         """
         all_wells = []
-        if type(op_data) is dict:
-            for k, v in op_data.items():
-                all_wells.append(self.get_wells(v))
-        elif type(op_data) is list:
+        if isinstance(op_data, dict):
+            for j, k in op_data.items():
```
sal-strateos, line 181: Would be good to know what j and k are here

autoprotocol/instruction.py
cj-bonilla: I was pylinting and they said to remove it but I agree it could open a can of worms 😓 I'll put it back haha

autoprotocol/instruction.py
cj-bonilla: this was more of a stylistic change. I personally don't like returning in if statements

autoprotocol/instruction.py
cj-bonilla: this...I do not know 😅 again, pylint was not happy with the original variables of a piece of code that was written before me. I'll look into what they are and change it!

autoprotocol/instruction.py
sal-strateos: it might be totally fine, but if you can't pass pylint I would get input and approval from someone more familiar with this repo

sal-strateos (approved):
lgtm!

jguiel (approved):
lets get this money 🚀

PR #383, opened 2023-02-07
SCIENG-175 Acoustic Transfer property overwrite fix and warning silencer

jguiel:
- Fixes bug of unrelated property deletion when propagating new properties via acoustic transfer.
- Allows the use of a kwarg in `AcousticTransfer.add_properties(self, properties, silence_warnings=False)` to silence the "overwriting property" warning.

### DIFF
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -71,7 +71,7 @@ class EntityPropertiesMixin:
         self.properties = properties.copy()
         return self
 
-    def add_properties(self, properties, silence_warnings=False):
+    def add_properties(self, properties):
         """
         Add properties to the properties attribute of an entity (ie: Container or Well).
 
@@ -83,8 +83,6 @@ class EntityPropertiesMixin:
         ----------
         properties : dict
             Dictionary of properties to add to an entity.
-        silence_warnings : bool
-            Silence warnings when overwriting existing properties
 
         Returns
         -------
@@ -98,12 +96,9 @@ class EntityPropertiesMixin:
             if key in self.properties:
                 if isinstance(current_value, list) and isinstance(new_value, list):
                     current_value.extend(new_value)
-                elif silence_warnings:
-                    self.properties[key] = new_value
                 else:
                     message = (
-                        f"Overwriting existing property {key} for {self}. To silence this warning, set"
-                        f" silence_warnings=True"
+                        f"Overwriting existing property {key} for {self}."
                     )
                     warnings.warn(message=message)
                     self.properties[key] = new_value
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -1,5 +1,3 @@
-# pylint: skip-file
-
 """
 Container, Well, WellGroup objects and associated functions
 
--- a/autoprotocol/container.py
+++ b/autoprotocol/container.py
@@ -71,7 +71,7 @@ class EntityPropertiesMixin:
         self.properties = properties.copy()
         return self
 
-    def add_properties(self, properties):
+    def add_properties(self, properties, silence_warnings=False):
         """
         Add properties to the properties attribute of an entity (ie: Container or Well).
 
@@ -82,19 +82,24 @@ class EntityPropertiesMixin:
         Parameters
         ----------
         properties : dict
-            Dictionary of properties to add to a entity.
+            Dictionary of properties to add to an entity.
+        silence_warnings : bool
+            Silence warnings when overwriting existing properties
 
         Returns
         -------
         self
             Container or Well with modified properties
         """
+
         self.validate_properties(properties)
         for key, new_value in properties.items():
             current_value = self.properties.get(key)
             if key in self.properties:
                 if isinstance(current_value, list) and isinstance(new_value, list):
                     current_value.extend(new_value)
+                elif silence_warnings:
+                    self.properties[key] = new_value
                 else:
                     message = f"Overwriting existing property {key} for {self}"
                     warnings.warn(message=message)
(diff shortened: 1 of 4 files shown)


autoprotocol/container.py
```diff
@@ -71,7 +71,7 @@ def set_properties(self, properties):
         self.properties = properties.copy()
         return self
 
-    def add_properties(self, properties):
+    def add_properties(self, properties, silence_warnings=False):
```
sal-strateos, line 74: I would make this more specific to the action. Something along the lines of "overwrite_existing_keys"

test/protocol_test.py
```diff
@@ -1432,8 +1432,11 @@ def test_propagate_properties(self, dummy_protocol):
         echo.well(0).set_volume("2:microliter")
         echo.well(0).set_properties({"test_well_property": True})
         dest.well(0).set_properties({"test_well_property": False})
+        dest.well(0).set_properties({"test_well_existing_property": True})
```
sal-strateos, line 1435: Should this be a test for property addition instead of setting?

sal-strateos (dismissed):
Discussed offline. Looks good to me

EribertoLopez (dismissed):
Do we really want to skip pylinting our most used files? The developer can import a dependency to silence warning from overwritting properties. What is the benefit of silencing the warning from a specific add_properties call if we are looping across wells and adding props?

EribertoLopez:
For the pylint issues you may be seeing you should not be running pylint on its own. You should be running the `pre-commit run --all-files` command which will point to the .pylintrc config file

autoprotocol/protocol.py
```diff
@@ -1677,7 +1679,7 @@ def acoustic_transfer(
             if v > Unit(0, "microliter"):
                 transfers.append(xfer)
             if self.propagate_properties:
-                d.set_properties(s.properties)
+                d.add_properties(s.properties, silence_warnings=True)
```
EribertoLopez, line 1682: The warning dependency is on line 12 in this module. Instead of changing the signature of the add_properties method, what about using `warnings.filterwarnings("ignore")` instead. If this is specific to acoustic_transfer and will be true every time then we should be able to do something like:
```python
if self.propagate_properties:
     warnings.filterwarnings("ignore")
     d.add_properties(s.properties)
```
Or you could put it in the application code to filter the warnings that are coming from your acoustic_transfer call
```python

def my_package_protocol(protocol, params):
    ...
    warnings.filterwarnings("ignore")
    protocol.acoustic_transfer(...)

**Comment by EribertoLopez** [comment]
If you do decide to to change the add_properties signature, then we will also want to make sure we add a test for Container.add_properties as well.

File: autoprotocol/protocol.py (hunk 3 above)
Comment (jguiel) [reply autoprotocol/protocol.py:1682]: That was my initial plan, but josh nor I could figure out how to test if it was _actually_ silencing the warnings. We spent too long on it, and this was just as effective. And wouldn't `filterwarnings` filter other possible warnings? This method just skips the one `warn` call

**Comment by jguiel** [comment]
@EribertoLopez I opted for skipping the pylint because there were +200 pylint errors for each file that I would need to fix. This file has never been properly linted, and many people apparently like using one-letter variable names 😏 If I were to lint, it would def be it's own PR. Since this isn't CI'ed by Jenkins, we weren't sure how to skip the lint of my branch.

**Comment by sal-strateos** [comment]
> @EribertoLopez I opted for skipping the pylint because there were +200 pylint errors for each file that I would need to fix. This file has never been properly linted, and many people apparently like using one-letter variable names 😏 If I were to lint, it would def be it's own PR. Since this isn't CI'ed by Jenkins, we weren't sure how to skip the lint of my branch.

I wonder if the errors you saw were due to not using the config file within this repo? I think this would explain the new errors you saw

**Comment by EribertoLopez** [comment]
> @EribertoLopez I opted for skipping the pylint because there were +200 pylint errors for each file that I would need to fix. This file has never been properly linted, and many people apparently like using one-letter variable names 😏 If I were to lint, it would def be it's own PR. Since this isn't CI'ed by Jenkins, we weren't sure how to skip the lint of my branch.

@jguiel We use Github actions for the CI in this repo you can see the CI / Lint stage in the checks. The CI uses the pre-commit command to run the lint tests. Sorry, but I can't let this pass if you want to skip linting whole files :/. The current master branch has no problems with running the command. Please make sure your local configuration is setup correctly by following https://github.com/autoprotocol/autoprotocol-python/blob/master/CONTRIBUTING.rst .

File: autoprotocol/protocol.py (hunk 3 above)
Comment (EribertoLopez) [reply autoprotocol/protocol.py:1682]: Use the https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings the warnings module has to catch the warnings from your package protocol code.
```python
def my_package_protocol(protocol, params):
    ...
    with warnings.catch_warnings():
         protocol.acoustic_transfer(...)
```

autoprotocol/protocol.py
```diff
@@ -1677,7 +1677,9 @@ def acoustic_transfer(
             if v > Unit(0, "microliter"):
                 transfers.append(xfer)
             if self.propagate_properties:
-                d.set_properties(s.properties)
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore")
+                    d.add_properties(s.properties)
```
joshhacksthings, line 1682: for now, let's only worry about adding properties. let's remove stuff around warnings, assuming our current implementations can be patched to silence these warnings if we encounter them

joshhacksthings (changes requested):

joshhacksthings (dismissed):
lgtm, just update changelog!

@EribertoLopez pinging you I've approved, any last comments, swoop in now

joshhacksthings (approved):

PR #388, opened 2023-02-27
serialize dispense column

joshhacksthings:

PR #389, opened 2023-02-28
fix(unit): serialization into string not dict

EribertoLopez:
#### Issue:
```
asdict(DispenseColumn) returns {'column': 1, 'volume': {'value': 700.0, 'units': 'microliter'}}
```
- Unit dataclass is being serialized into a dict but is expected as a str

#### Solution:
- Adding `asdict(obj, dict_factory=unit_as_strings_factory)`


test/builder_test.py
```diff
@@ -58,6 +59,11 @@ class TestDispenseBuilders(object):
     def test_column(self):
         for column in self.columns_reference:
             assert DispenseColumn(**column) == Dispense.builders.column(**column)
+            assert (
+                asdict(DispenseColumn(**column))
+                == asdict(Dispense.builders.column(**column))
+                == {"column": column["column"], "volume": str(column["volume"])}
+            )
```
joshhacksthings, line 66: nice, we're asserting that DispenseColumn serializes properly here, but should we also include something like this in a test for the Dispense Instruction? 

Currently, [my PR](https://github.com/autoprotocol/autoprotocol-python/pull/388) makes the change that calls asdict on DispenseColumn inside the dispense method.

I guess I could add my commit to this branch? Just LMK!

joshhacksthings (approved):
lgtm!

PR #391, opened 2023-03-25
Revert rec tip type assignment

sal-strateos:
Identified an error scheduling runs that contained liquid transfer instructions. This reverts a recently introduced method.

Previous behavior called `_rec_tip_type` from the parent class of `Transfer` but didn't assign it to self.tip_type

Scheduled a run and confirmed this fixes the error.

### DIFF
--- a/autoprotocol/liquid_handle/transfer.py
+++ b/autoprotocol/liquid_handle/transfer.py
@@ -102,10 +102,6 @@ class Transfer(LiquidHandleMethod):
         self._source_liquid = None
         self._destination_liquid = None
 
-    def _rec_tip_type(self, volume: Unit):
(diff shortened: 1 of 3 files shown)


joshhacksthings (dismissed):
looks good! can you update changelog?

if you're intending on releasing immediately, make sure to add a version to changelog and `version.py`!

sal-strateos:
> looks good! can you update changelog?
> 
> if you're intending on releasing immediately, make sure to add a version to changelog and `version.py`!

Merging based on consensus, but future us should keep in mind that using this would improve liquid transfers.

cj-bonilla (approved):
lgtm 🪨

PR #393, opened 2023-03-30
SCIENG-223: Update dependencies to allow for mantis liquid handle dispense

cj-bonilla:
[SCIENG-223](https://strateos.atlassian.net/browse/SCIENG-223)

Started new branch, but from `master`

[SCIENG-223]: https://strateos.atlassian.net/browse/SCIENG-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

referencing [#1109](https://github.com/strateos/internal_protocols/pull/1109)

### Test New Txappy
Release `txappy` and then:
- [x] add a new test that uses a tip type as a source
- [x] [submit to WEB](https://webapp.staging.strateos.com/transcriptic/p1ax2tm4re3b9pezp/runs/r1ax569wg75kvj2v8/instructions)

Testing Mantis Driver Task will cover following tests
### Should pass
- [ ] [Test `device_mode_params` 3 different levels of param specification](https://webapp.staging.strateos.com/transcriptic/runspage/queue/accepted/runs/r1ax35hdyp5yjh3ru/prime)`
- [ ] [Test `dispense` to full 96-plate](https://webapp.staging.strateos.com/transcriptic/runspage/queue/accepted/runs/r1ax36ujrgrmjdrxb/prime)
- [ ] [Test `dispense` from many sources to 1 destination _over several instructions_](https://webapp.staging.strateos.com/transcriptic/runspage/queue/accepted/runs/r1ax372yjeysb2fyp/prime)
### Should fail
- [ ] [Test `dispense` from 1 source to 2+ destination plates](https://webapp.staging.strateos.com/transcriptic/runspage/queue/accepted/runs/r1ax37adavd7uc9em/prime)
- [ ] [Test `dispense` from many source to 1 destination](https://webapp.staging.strateos.com/transcriptic/runspage/queue/accepted/runs/r1ax376fzjgzvvp8k/prime)

### DIFF
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1536,7 +1536,7 @@ class LiquidHandleBuilders(InstructionBuilders):
         chip_material: Optional[str] = None,
         nozzle: Optional[str] = None,
         diaphragm: Optional[int] = None,
-        nozzle_size: Optional[Unit] or Optional[str] = None,
+        nozzle_size: Optional[Union[Unit, str]] = None,
         tubing: Optional[str] = None,
         z_drop: Optional[Unit] or Optional[str] = None,
         viscosity: Optional[str] = None,
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1539,7 +1539,7 @@ class LiquidHandleBuilders(InstructionBuilders):
         nozzle_size: Optional[Unit] = None,
         tubing: Optional[str] = None,
         z_drop: Optional[Unit] = None,
-        viscosity: Optional[str] = None
+        viscosity: Optional[str] = None,
     ):
         """Helper for building device level mode_params
 
@@ -1590,14 +1590,14 @@ class LiquidHandleBuilders(InstructionBuilders):
             If chip_material is not None or in the allowable list of tempest chip materials
         """
         params: dict = {
-           "model": model,
-           "material": chip_material,
-           "nozzle": nozzle,
-           "diaphragm": diaphragm,
-           "nozzle_size": nozzle_size,
-           "tubing": tubing,
-           "z_drop": z_drop,
-           "viscosity": viscosity
+            "model": model,
+            "material": chip_material,
+            "nozzle": nozzle,
+            "diaphragm": diaphragm,
+            "nozzle_size": nozzle_size,
+            "tubing": tubing,
+            "z_drop": z_drop,
+            "viscosity": viscosity,
         }
         if device not in ["x_mantis", "x_tempest_chip"]:
             raise ValueError(
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1568,11 +1568,6 @@ class LiquidHandleBuilders(InstructionBuilders):
         viscosity: Optional[String]
             one of "1", "2-5", "6-10", "11-20", "21-25"
 
-        Returns
-        -------
-        dict
-            device_mode_params for a LiquidHandle instruction
-
         Raises
         ------
         ValueError
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1568,6 +1568,11 @@ class LiquidHandleBuilders(InstructionBuilders):
         viscosity: Optional[String]
             one of "1", "2-5", "6-10", "11-20", "21-25"
 
+        Returns
+        -------
+        dict
+            device_mode_params for a LiquidHandle instruction
+
         Raises
         ------
         ValueError
@@ -1598,7 +1603,7 @@ class LiquidHandleBuilders(InstructionBuilders):
         return {device: device_dict}
 
     @staticmethod
-    def validate_device_params(device: str, device_dict: dict):
+    def validate_device_params(device: str, device_dict: dict) -> None:
         """
         Helper validation function to validate device liquid handling params
         Parameters
@@ -1607,11 +1612,6 @@ class LiquidHandleBuilders(InstructionBuilders):
             either x_mantis or x_tempest_chip
         device_dict: dict
             Dictionary of all device params, as seen in function device_mode_params
-
-        Returns
-        -------
-        dict
-            default_dict for the specified device
 
         Raises
         ------
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1548,13 +1548,13 @@ class LiquidHandleBuilders(InstructionBuilders):
 
         Parameters
         ----------
-        device: string
+        device: str
             x_tempest_chip or x_mantis
-        model: string, optional
+        model: Optional[str]
             Tempest chip or mantis model.
-        chip_material: string, optional
+        chip_material: Optional[str]
             Material that the tempest chip is made of.
-        nozzle: string, optional
+        nozzle: Optional[str]
             Type of chip nozzle for tempest.
         diaphragm: Optional[Integer]
             any integer between 0 and 100, inclusive
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1562,6 +1562,7 @@ class LiquidHandleBuilders(InstructionBuilders):
             one of "0.1:mm", "0.2:mm", "0.5:mm"
         tubing: Optional[String]
             one of "LV", "HV", "P200", "P1000"
+            TODO: support that source is pipette-tip type
         z_drop: Optional[Unit[Length]]
             within the range: 0:mm - 100:mm, inclusive
         viscosity: Optional[String]
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1666,7 +1666,10 @@ class LiquidHandleBuilders(InstructionBuilders):
                         compare_val = Unit(value)
                     else:
                         compare_val = value
-                    if accepted_range[0] > compare_val > accepted_range[1]:
+                    if (
+                        compare_val < accepted_range[0]
+                        or compare_val > accepted_range[1]
+                    ):
                         error_values.update({key: value})
                 # The following logic should be able to support container_type.shortname
                 elif key == "tubing":
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1589,6 +1589,16 @@ class LiquidHandleBuilders(InstructionBuilders):
         ValueError
             If chip_material is not None or in the allowable list of tempest chip materials
         """
+        params: dict = {
+           "model": model,
+           "material": chip_material,
+           "nozzle": nozzle,
+           "diaphragm": diaphragm,
+           "nozzle_size": nozzle_size,
+           "tubing": tubing,
+           "z_drop": z_drop,
+           "viscosity": viscosity
+        }
         if device not in ["x_mantis", "x_tempest_chip"]:
             raise ValueError(
                 f"Device is {device}. It must be: [x_tempest_chip, x_mantis]"
@@ -1618,26 +1628,10 @@ class LiquidHandleBuilders(InstructionBuilders):
                 "viscosity": None,
             }
         device_dict: dict = {}
-        if any([
-            model,
-            chip_material,
-            nozzle,
-            diaphragm,
-            nozzle_size,
-            tubing,
-            z_drop,
-            viscosity
-        ]):
-            device_dict.update({
-                "model": model or default_params["model"],
-                "material": chip_material or default_params["material"],
-                "nozzle": nozzle or default_params["nozzle"],
-                "diaphragm": diaphragm or default_params["diaphragm"],
-                "nozzle_size": nozzle_size or default_params["nozzle_size"],
-                "tubing": tubing or default_params["tubing"],
-                "z_drop": z_drop or default_params["z_drop"],
-                "viscosity": viscosity or default_params["viscosity"],
-            })
+        if any(params.values()):
+            for key, value in params.items():
+                device_dict.update({key: value or default_params[key]})
+
         LiquidHandleBuilders.validate_device_params(device, device_dict)
 
         return {device: device_dict}
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1603,41 +1603,21 @@ class LiquidHandleBuilders(InstructionBuilders):
             raise ValueError(
                 f"Device is {device}. It must be: [x_tempest_chip, x_mantis]"
             )
-        if device == "x_mantis":
-            # If device is mantis, set default params
-            default_params: dict = {
-                "model": "high_volume",
-                "material": None,
-                "nozzle": None,
-                "diaphragm": 0,
-                "nozzle_size": "0.1:mm",
-                "tubing": "LV",
-                "z_drop": "0.0:mm",
-                "viscosity": "1",
-            }
-        else:
-            # Otherwise, default to tempest default params
-            default_params: dict = {
-                "model": "high_volume",
-                "material": "pfe",
-                "nozzle": "standard",
-                "diaphragm": None,
-                "nozzle_size": None,
-                "tubing": None,
-                "z_drop": None,
-                "viscosity": None,
-            }
+
+        default_params = LiquidHandleBuilders.validate_device_params(device, params)
+
         device_dict: dict = {}
         if any(params.values()):
             for key, value in params.items():
-                device_dict.update({key: value or default_params[key]})
-
-        LiquidHandleBuilders.validate_device_params(device, device_dict)
+                if key not in default_params:
+                    device_dict.update({key: value or None})
+                else:
+                    device_dict.update({key: value or default_params[key]})
 
         return {device: device_dict}
 
     @staticmethod
-    def validate_device_params(device: str, device_dict: dict) -> None:
+    def validate_device_params(device: str, device_dict: dict) -> dict:
         """
         Helper validation function to validate device liquid handling params
         Parameters
@@ -1669,7 +1649,7 @@ class LiquidHandleBuilders(InstructionBuilders):
             accepted_params: dict = {
                 "model": ["high_volume"],
                 "nozzle": ["standard"],
-                "material": ["silicone", "pfe"],
+                "material": ["pfe", "silicone"],
             }
         else:
             raise ValueError(
@@ -1691,6 +1671,12 @@ class LiquidHandleBuilders(InstructionBuilders):
             raise ValueError(
                 f"Incorrect params: {error_values}. It must be {accepted_params}"
             )
+
+        default_dict: dict = {}
+        for key, value in accepted_params.items():
+            default_dict.update({key: value[0]})
+
+        return default_dict
 
     @staticmethod
     def move_rate(
--- a/autoprotocol/builders.py
+++ b/autoprotocol/builders.py
@@ -1582,12 +1582,6 @@ class LiquidHandleBuilders(InstructionBuilders):
         ------
         ValueError
             If device is not of the following: [x_tempest_chip, x_mantis]
-        ValueError
-            If model is not None or "high_volume"
-        ValueError
-            If nozzle is not None or "standard"
-        ValueError
-            If chip_material is not None or in the allowable list of tempest chip materials
         """
         params: dict = {
             "model": model,
@@ -1673,6 +1667,7 @@ class LiquidHandleBuilders(InstructionBuilders):
             )
 
         default_dict: dict = {}
+        # Default to the first value in the accepted params dict
         for key, value in accepted_params.items():
             default_dict.update({key: value[0]})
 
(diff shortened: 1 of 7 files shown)


test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +856,166 @@ def test_tempest_bad_chip_param(self):
...
+                "model": "high_volume",
+                "diaphragm": 0,
+                "nozzle_size": "0.1:mm",
+                "tubing": "LV",
+                "z_drop": "0.0:mm",
+                "viscosity": "1",
+            }
+        }
+
+        assert instruction.data["mode_params"] == mode_params
```
cj-bonilla, line 895: From PR 390: 

_can we also make assertions here about the shape of the LiHa? ie how many rows and columns would we dispense into with these params?_

cj-bonilla: _similar note here, let's assert we have 1 row and 1 column in our mode params. and as a follow-up question, is it safe to assume/assert in the dispense method that if x_mantis is specified that we also default to 1 row and 1 column?_

cj-bonilla: Added liha shape assertion to test in latest commit!

autoprotocol/builders.py
```diff
@@ -1542,12 +1548,30 @@ def device_mode_params(
...
+            one of "LV", "HV", "P200", "P1000"
+            TODO: How do we want to handle these tip types?
+            In the future, we're going to have tempest source support
+            this new tip type container type which will get a shortname
+            akin to `pipette-tip-p200` or something. If we align these
+            two values to that shortname, we can just pass it in via
+            `source_container.container_type.shortname`
+        z_drop: Optional[Unit[Length]]
+            within the range: 0:mm - 100:mm, inclusive
+        viscosity: Optional[String]
```
joshhacksthings, line 1573: this is a super nit, but the type descriptions for these are a different style than the ones above

ie, `nozzle: string, optional` vs `diaphragm: Optional[Integer]`

we should align on one format for all

autoprotocol/builders.py
```diff
@@ -1542,12 +1548,30 @@ def device_mode_params(
...
+        nozzle_size: Optional[Unit[Length]]
+            one of "0.1:mm", "0.2:mm", "0.5:mm"
+        tubing: Optional[String]
+            one of "LV", "HV", "P200", "P1000"
+            TODO: How do we want to handle these tip types?
+            In the future, we're going to have tempest source support
+            this new tip type container type which will get a shortname
+            akin to `pipette-tip-p200` or something. If we align these
+            two values to that shortname, we can just pass it in via
+            `source_container.container_type.shortname`
```
joshhacksthings, line 1570: do we still need this note? 

i guess we're leaving this decision to when we implement this tip types?

autoprotocol/builders.py
```diff
@@ -1557,38 +1581,101 @@ def device_mode_params(
...
+        }
+        if device not in ["x_mantis", "x_tempest_chip"]:
+            raise ValueError(
+                f"Device is {device}. It must be: [x_tempest_chip, x_mantis]"
+            )
+
+        default_params = LiquidHandleBuilders.validate_device_params(device, params)
+
+        device_dict: dict = {}
+        if any(params.values()):
```
joshhacksthings, line 1604: so, if there are any values, we iterate through and assign _something_ to each key

but what if there are no values? we just pass an empty dict?
should we do similar behavior for any key that doesn't have a value?

autoprotocol/protocol.py
```diff
@@ -1185,6 +1207,8 @@ def liquid_handle_dispense(
         default_num_dispense_chips_in_source: int = 1
         default_max_num_dispense_chips: int = 12
         remaining_num_chips_to_specify = default_max_num_dispense_chips
+        if device == "x_mantis":
+            rows = 1
```
joshhacksthings, line 1211: should we assign this for the protocol writer or should we validate rowx,columns vs device and like error at them?

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +861,170 @@ def test_tempest_bad_chip_param(self):
                 nozzle="abc",
             )
 
+    def test_mantis_default_params(self):
+        """Tests mantis default params"""
```
joshhacksthings, line 865: are these mantis default params or are they device-agnostic params? like if we dont have a `mode_params` dict, are they even device specific?

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +861,170 @@ def test_tempest_bad_chip_param(self):
...
+        )
+
+        assert "mode_params" not in instruction.data
+
+        instruction = self.protocol.liquid_handle_dispense(
+            source=self.tube.well(0),
+            destination=self.flat.well(0),
+            volume="5:uL",
+            liquid=ProteinBuffer,
+            device="x_mantis",
```
joshhacksthings, line 880: it might seem a bit verbose, but i think we should include rows and columns explicitly here, instead of relying on the method to set it

joshhacksthings (changes requested):
it's getting there. left a few commnets.

one thing i didnt leave in comments but want addressed is how should we handle under-specified mode params?

ie if i only specify z_drop and viscosity, should we be assigning defaults in this code? My gut is that we dont pass anything. no `nozzle` value in the dict if `nozzle` isnt provided. thoughts? if we go down this route can you include a test that demosntrates this?

test/liquid_handle_dispense_test.py
cj-bonilla: ok!

autoprotocol/protocol.py
cj-bonilla: I usually lean towards auto-assigning based on the capabilities of the device, but if we want to give more flexibility, I'm fine with a validation & erroring at them!

autoprotocol/builders.py
cj-bonilla: on it!

autoprotocol/builders.py
cj-bonilla: I do not know if we still need this? Do we plan to do this? Or should we just nix it for now?

autoprotocol/protocol.py
```diff
@@ -1469,7 +1491,15 @@ def equalize_lengths(
             # Update source volume
             source_location.add_volume(-total_volume_dispensed)
         device_mode_params = LiquidHandleBuilders.device_mode_params(
-            model=model, chip_material=chip_material, nozzle=nozzle
+            device=device or "x_tempest_chip",
```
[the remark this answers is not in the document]
cj-bonilla: ahhh yes, that makes sense, but i guess cuz I was auto-assigning based on the selected device. I can switch for more flexibility!

test/liquid_handle_dispense_test.py
cj-bonilla: these are mantis default params! If we don't have a `mode_params` dict, it errors out!

autoprotocol/protocol.py
cj-bonilla: I ended up just giving the protocol writer that flexibility and adding in a validation in `util.py`. Should be in the commit just now!

autoprotocol/builders.py
cj-bonilla: I may have misunderstood this note you left in PR#390 ? Not sure if you meant for me to work on this...:sweat-smile:

autoprotocol/builders.py
cj-bonilla: Yeah, depending on the device, there's a default device dict! Right now it's defaulted to `x_tempest_chip`, but I can set it to a required arg and make it backwards-compatible!

cj-bonilla: This is more so aligned with the original code; if we want to redo that logic, then I feel like that might be a much larger lift than expected 😭 I am more than happy to work on it, just also want to make sure it's within our scope

autoprotocol/builders.py
cj-bonilla: I added in new logic in the most recent commit to possibly support this. lmk what you think

autoprotocol/protocol.py
```diff
@@ -1083,6 +1092,20 @@ def liquid_handle_dispense(
             Tempest nozzle type, currently only support "standard".
             The three chip parameters: model, chip_material, and nozzle will be
             used in liquid handle mode_params to allow tempest chip specification.
+        diaphragm : int, optional
```
joshhacksthings, line 1095:
```suggestion
        diaphragm : Optional[int]
```

autoprotocol/protocol.py
```diff
@@ -1030,9 +1031,15 @@ def liquid_handle_dispense(
         columns: int = 1,
         method: DispenseMethod = DispenseMethod,
         liquid: LiquidClass = LiquidClass,
+        device: Optional[str] = "x_tempest_chip",
         model: Optional[str] = None,
         chip_material: Optional[str] = None,
         nozzle: Optional[str] = None,
+        diaphragm: Optional[int] = None,
+        nozzle_size: Optional[Unit] = None,
```
joshhacksthings, line 1039: technically we accept Optional Unit or str 🤔

autoprotocol/util.py
```diff
@@ -213,3 +213,22 @@ def _check_container_type_with_shape(container_type, shape):
             f"a full row consists of {format_columns} columns and a full "
             f"column consists of {format_rows} rows, but {shape} was specified."
         )
+
+
+def _validate_liha_shape(device: str, shape: dict) -> None:
+    """Validates LiHa shape for liquid handle dispense"""
```
joshhacksthings, line 219: i got chatgpt to make a docstring for us. can you edit and include this?

```suggestion
    """Validates LiHa shape for liquid handle dispense.

    Args:
        device (str): The LiHa device name. Currently supported devices are 
                      "x_mantis" and "x_tempest_chip".
        shape (dict): A dictionary containing the number of rows and columns
                      for the LiHa shape. For example, {"rows": 8, "columns": 1}.

    Raises:
        ValueError: If the provided device name is not one of the supported 
                    devices or if the provided shape is not valid for the 
                    specified device.

    Returns:
        None

    Example:
        >>> _validate_liha_shape("x_mantis", {"rows": 1, "columns": 1})
        None
    """
```

autoprotocol/util.py
```diff
@@ -213,3 +213,22 @@ def _check_container_type_with_shape(container_type, shape):
             f"a full row consists of {format_columns} columns and a full "
             f"column consists of {format_rows} rows, but {shape} was specified."
         )
+
+
+def _validate_liha_shape(device: str, shape: dict) -> None:
```
joshhacksthings, line 218: can you write a test for this?

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +858,170 @@ def test_tempest_bad_chip_param(self):
...
+            }
+        }
+
+        liha_shape = instruction.data["shape"]
+        assert liha_shape["rows"] == 1 and liha_shape["columns"] == 1
+        assert instruction.data["mode_params"] == mode_params
+
+    def test_mantis_bad_params(self):
+        """Tests mantis bad params"""
+        # Incorrect model param and incorrect liha shape
```
joshhacksthings, line 955: why does this one fail?

joshhacksthings: is this just testing that `x_mantis` without rows/columns fails?

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +858,170 @@ def test_tempest_bad_chip_param(self):
...
+                self.protocol.liquid_handle_dispense(
+                    source=self.tube.well(0),
+                    destination=self.flat.well(0),
+                    volume="5:uL",
+                    rows=1,
+                    columns=1,
+                    liquid=ProteinBuffer,
+                    device="x_mantis",
+                    viscosity="100",
+                )
```
joshhacksthings, line 1023: can you add one more bad test that asserts we can't pass in `x_mantis` with a tempest-specific param?

joshhacksthings (changes requested):
getting close!

here's a summary of things i noted
- [ ] update docstring for inputs that accept Unit or str
- [ ] update validation for these inputs that accept Unit or str
- [ ] add test for new liha shape util function
- [ ] add test demonstrating bad inputs for dispense with `x_mantis` and tempest specific param

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,96 @@ def device_mode_params(
...
-        """
-        # tempest chip specification
-        if (model != "high_volume") and (model is not None):
-            raise ValueError(f"Tempest chip model is {model}. It must be: high_volume.")
-        if (nozzle != "standard") and (nozzle is not None):
-            raise ValueError(f"Tempest nozzle is {nozzle}. It must be: standard.")
-        if (chip_material not in ["silicone", "pfe"]) and (chip_material is not None):
+            If device is not of the following: [x_tempest_chip, x_mantis]
+        """
+        params: OrderedDict = {
```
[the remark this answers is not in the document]
cj-bonilla: oh rip LOL ok on it!

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +858,170 @@ def test_tempest_bad_chip_param(self):
...
+            source=self.tube.well(0),
+            destination=self.flat.well(0),
+            volume="5:uL",
+            rows=1,
+            columns=1,
+            liquid=ProteinBuffer,
+            device="x_mantis",
+            model="high_volume",
+            diaphragm=0,
+            nozzle_size="0.1:mm",
```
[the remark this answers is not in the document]
cj-bonilla: It did not like it lol. So I changed things to unit?

autoprotocol/protocol.py
cj-bonilla: AGH ok yeah I think I resolved it? The output will always spit out a `Unit` type _but_ accepts both string and unit. I'll do the same for `z_drop`

autoprotocol/util.py
cj-bonilla: LOL love, committing this suggestion

test/liquid_handle_dispense_test.py
cj-bonilla: this one fails because it doesn't send a `row`/`col`, which defaults to `tempest` shape, _but_ I specified the `device` as `mantis`

cj-bonilla: I'll add in the comment that it's to test the incorrect liha shape for the mantis

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,96 @@ def device_mode_params(
...
-                model = "high_volume"
-            if chip_material is None:
-                chip_material = "pfe"
-            if nozzle is None:
-                nozzle = "standard"
-            x_tempest_chip = {
-                "model": model,
-                "material": chip_material,
-                "nozzle": nozzle,
+        LiquidHandleBuilders.validate_device_params(device, params)
```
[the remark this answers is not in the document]
cj-bonilla: ah yeah if it's empty let's just not validate it

autoprotocol/util.py
cj-bonilla: on it like blue bonnet

test/liquid_handle_dispense_test.py
cj-bonilla: done!

test/liquid_handle_dispense_test.py
```diff
@@ -856,6 +858,170 @@ def test_tempest_bad_chip_param(self):
                 nozzle="abc",
             )
 
+    def test_mantis_default_params(self):
```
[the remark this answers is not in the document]
cj-bonilla: done!

autoprotocol/builders.py
```diff
@@ -1531,9 +1531,15 @@ def mode_params(
 
     @staticmethod
     def device_mode_params(
+        device: str,
         model: Optional[str] = None,
         chip_material: Optional[str] = None,
         nozzle: Optional[str] = None,
+        diaphragm: Optional[int] = None,
+        nozzle_size: Optional[Unit] or Optional[str] = None,
```
joshhacksthings, line 1539: you want the `Union` type

```suggestion
        nozzle_size: Optional[Union[Unit, str]] = None,
```

autoprotocol/builders.py
```diff
@@ -1531,9 +1531,15 @@ def mode_params(
...
+        tubing: Optional[str] = None,
+        z_drop: Optional[Unit] or Optional[str] = None,
```
joshhacksthings, line 1541:
```suggestion
        z_drop: Optional[Union[Unit, str]] = None,
```

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,127 @@ def device_mode_params(
...
+                "tubing": tubing,
+                "z_drop": z_drop,
+                "viscosity": viscosity,
+            }
+        )
+        if device not in ["x_mantis", "x_tempest_chip"]:
+            raise ValueError(
+                f"Device is {device}. It must be: [x_tempest_chip, x_mantis]"
+            )
+        if any(params.values()):
```
joshhacksthings, line 1597: what do you think about nesting lines 1600-1607 under this, too? my thinking here is like we don't need to traverse this dict if there are no parameters set?

we'd have to pull `default_dict` out i guess, so we could return it

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,127 @@ def device_mode_params(
...
+            params for the input device
+        KeyError
+            If input device_dict key is not in accepted_params keys
+        """
+        if device == "x_mantis":
+            # If device is mantis, set accepted params
+            accepted_params: dict = {
+                "model": ["high_volume", "low_volume"],
+                "diaphragm": [0, 100],
+                "nozzle_size": [Unit("0.1:mm"), Unit("0.2:mm"), Unit("0.5:mm")],
```
joshhacksthings, line 1636: butbutbutbut what if the user specifies `nozzle_size` as string?!

i think what we should do, before comparing to this dict, is call `parse_unit(NOZZLE_SIZE)` on the user-specified nozzle_size... and `z_drop`

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,127 @@ def device_mode_params(
...
+                elif key == "z_drop":
+                    accepted_range: List = list(accepted_params[key])
+                    if isinstance(value, str):
+                        compare_val = Unit(value)
+                    else:
+                        compare_val = value
+                    if (
+                        compare_val < accepted_range[0]
+                        or compare_val > accepted_range[1]
+                    ):
```
joshhacksthings, line 1672: i _think_ you could write this like

```suggestion
                    if accepted_range[0] > compare_val > accepted_range[1]:
```

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,127 @@ def device_mode_params(
...
+                        error_values.update({key: value})
+                # The following logic should be able to support container_type.shortname
+                elif key == "tubing":
+                    accepted_val: bool = False
+                    for tubing_val in accepted_params[key]:
+                        if tubing_val in value.upper():
```
joshhacksthings, line 1678: we can totally deal with this later, but i dont think we should be calling `upper` on this as it will break. when we eventually support the `pipette-tip-p200` or whatever shortname

autoprotocol/protocol.py
```diff
@@ -1030,9 +1031,15 @@ def liquid_handle_dispense(
...
+        tubing: Optional[str] = None,
+        z_drop: Optional[Unit] = None,
```
joshhacksthings, line 1041:
```suggestion
        z_drop: Optional[Union[Unit, str]] = None,
```

autoprotocol/protocol.py
joshhacksthings, line 1039:
```suggestion
        nozzle_size: Optional[Union[Unit, str]] = None,
```

autoprotocol/builders.py
```diff
@@ -1557,38 +1576,127 @@ def device_mode_params(
...
+        KeyError
+            If input device_dict key is not in accepted_params keys
+        """
+        if device == "x_mantis":
+            # If device is mantis, set accepted params
+            accepted_params: dict = {
+                "model": ["high_volume", "low_volume"],
+                "diaphragm": [0, 100],
+                "nozzle_size": [Unit("0.1:mm"), Unit("0.2:mm"), Unit("0.5:mm")],
+                "tubing": ["LV", "HV", "P200", "P1000"],
```
joshhacksthings, line 1637: how does your pytest that passes in `tubing = "pipette-tip-p200" work if that string isnt in this list?

joshhacksthings (approved):
i've approved this but take a look at all of my comments

autoprotocol/builders.py
cj-bonilla: ahh yeah ok makes sense!

joshhacksthings (approved):
lgtm!

PR #394, opened 2023-04-03
feat: [PP-12615] added missing informatics if provision volume greater than 900 microliter

jayanth-1:
Jira: https://strateos.atlassian.net/browse/PP-12615
Added missing informatics if provision volume greater than 900 microliter

**Test plan**
Added unit test cases to verify it


gokulatzemoso (commented):
Mostly looks good, let's test provision informatics where volume < 900ul as well?

gokulatzemoso (dismissed):

test/test_provision_mixture_informatics.py
joshhacksthings, line 1: how would you feel about creating this method in https://github.com/autoprotocol/autoprotocol-python/blob/master/test/test_util.py? or even somewhere like `conftest.py`

This class isn't testing anything and I think it would make sense to keep our test helpers consolidated somewhere

joshhacksthings (changes requested):
looks good, just had some comments on testing patterns


from the ticket description, it looks like we are assigning informatics information when provisioning volume is > 900uL. 

This PR does assign informatics information when this volume is greater than 900uL, but I was wondering where this issue was reported from? Is this for autoprotocol's `protocol.provision` or txappy's `protocol.provision`? The latter is capable of provisioning volumes larger than 900uL, while the former can only provision volumes broken up into 900uL increments.

Here is the code for [human_provision](https://github.com/strateos/txappy/blob/3f7fe53d28fe4e339719eaa8a40d450e46930143/txappy/protocol.py#L1503)

jayanth-1:
> looks good, just had some comments on testing patterns
> 
> from the ticket description, it looks like we are assigning informatics information when provisioning volume is > 900uL.
> 
> This PR does assign informatics information when this volume is greater than 900uL, but I was wondering where this issue was reported from? Is this for autoprotocol's `protocol.provision` or txappy's `protocol.provision`? The latter is capable of provisioning volumes larger than 900uL, while the former can only provision volumes broken up into 900uL increments.
> 
> Here is the code for [human_provision](https://github.com/strateos/txappy/blob/3f7fe53d28fe4e339719eaa8a40d450e46930143/txappy/protocol.py#L1503)

We are not using txappy human provision, everywhere we are using autoprotocol provision. Please check [here](https://github.com/strateos/autoprotocol-service/blob/master/autoprotocol_service/nodes/provision_mixture/v0_0_2.py#L320) and in human provision, there is no informatics at all

joshhacksthings (dismissed):
thanks for the update to the tests Jayanth! Looks great!

PR #395, opened 2023-04-18
PP-12831

sabrinamohabbat:
This PR updates protocol.py to not pass in mode params and tip type in the liquid handle transfer instruction. 

Tested by generating a[ test run](https://secure.strateos.com/strateos_sd_qc/p1gxgkdddyg6sz/runs/r1axj4ca7qj9b462g/instructions) and you can see the last two liquid handle instructions dont have the mode params and tip type appended.

### DIFF
--- a/autoprotocol/protocol.py
+++ b/autoprotocol/protocol.py
@@ -7896,7 +7896,7 @@ class Protocol:
 
         # apply tip types to transfer methods
         for vol, met in zip(volume, method):
-            if not met.tip_type:
+            if met._has_calibration() and not met.tip_type:
                 try:
                     # met.tip_type = met._rec_tip_type(vol)
                     met._rec_tip_type(vol)
--- a/autoprotocol/version.py
+++ b/autoprotocol/version.py
@@ -1,2 +1,2 @@
 """Maintains current version of package"""
-__version__ = "10.2.3"
+__version__ = "10.2.4"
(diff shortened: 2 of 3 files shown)


joshhacksthings (approved):

autoprotocol/protocol.py
```diff
@@ -7896,7 +7896,7 @@ def informatics_helper(informatics, dest, multi_src):
 
         # apply tip types to transfer methods
         for vol, met in zip(volume, method):
-            if not met.tip_type:
+            if met._has_calibration() and not met.tip_type:
```
sal-strateos, line 7899: Curious, prior to this change, was a runtime error being raised?

joshhacksthings: Yep! Removing this is what introduced us setting filtered1000s in `mode_params`.

sal-strateos: The reason I ask is because there are tests in Txappy that depend on a runtime error being triggered so if we gate this with the `_has_calibration() method we may break the tests in Txappy. Here is the line to confirm and update if this is True: https://github.com/strateos/txappy/blob/751890f8a6353d385bc35177490d46cd4a1b7372/test/liquid_handle_test.py#L102

sabrinamohabbat: Will address the txappy tests and update them in a different PR.

